PR #28, opened 2017-03-23
Add Travis CI

saicheems:


.travis.yml
```diff
@@ -0,0 +1,11 @@
+sudo: required
+
+services:
+    - docker
+
+script:
+    - export GOPATH=${HOME}/gopath
```
vchudnov-g: Whose `$HOME` is this? Travis gets an account set up and this is its homedir? Is it persistent?

saicheems: It's the Travis instance's `$HOME`

There are some guaranteed default env variables, some of which I use in this script:
https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

PR #29, opened 2017-03-27
Support embedded nodes when parsing PyDoc

saicheems:


saicheems:
@tcoffee-google 

No, using `node.Data` didn't work

Admittedly, I wasn't really motivated enough to find out why. I tried it again to get the error:
`parselib_test.go:137: method signature has unexpected form: code`

tcoffee-google:
Oh, I understand --- node.Data gives you the contents of the `<code>` node, which is `code`. Present implementation is slightly odd, but seems correct and soon to be deprecated. LGTM.

PR #30, opened 2017-03-28
Fix Ruby compilecheck

saicheems:
Add support for cases where the init section of a snippet ends with a
`fetch_all` with no parameters.


src/snippetgen/compilecheck/internal/ruby/ruby.go
```diff
@@ -26,7 +26,7 @@ var (
 
 	// performCallRegex matches the first line of the method call.
 	// ex: "result = service.foo_bar()"
```
tcoffee-google: Please update comment.

saicheems: Done

PR #35, opened 2017-04-13
Add mockgen scripts

saicheems:
This commit introduces two scripts:

1. gen_mock.py

This script is responsible for generating a mock server for each method
in the given discovery doc.

2. gen_ovr.py

This script is responsible for generating an override file for the given
discovery doc that ensures all placeholders in the generated samples are
valid.


saicheems:
Hey guys, these scripts are 95% of the way there, just missing some docstrings. I'm comfortable with starting review.

The first one generates the mock servers from a discovery doc, and the second generates an overrides file which replaces the default string for all fields which are required and have a non-path-template pattern. It uses the `rstr` package to generate a random string that matches against a regex. There was another package I wanted to use which would generate the first string that matches against a regex... but the license was AGPL, which is not allowed at Google :(

saicheems:
Yes, let me do that first, and I'll re-tag you guys...

tcoffee-google:
Is this ready for review?

saicheems:
No, almost ready, I've just been updating it as I progress.

PR #36, opened 2017-04-14
WIP: Automate client library updates

tcoffee-google:
Currently includes Node.js and Ruby. Remains to be tested.

### DIFF
--- a/src/main/nodejs/nodejs.go
+++ b/src/main/nodejs/nodejs.go
@@ -1,0 +1,195 @@
+package nodejs
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"regexp"
+	"strings"
+	"sync"
+	"time"
+
+	"discovery-artifact-manager/common/errorlist"
+	"discovery-artifact-manager/main/common"
+)
+
+const (
+	logFile   = "CHANGELOG.md"
+	logUpdate = `##### %s - %s
+
+- Regenerated
+
+`
+)
+
+// updateLog updates the change log file on the given path with a new version entry, and returns the version number.
+func updateLog(path string) (string, error) {
+	return common.UpdateFile(path, logFile, func(log []byte) (changed []byte, bumped string, err error) {
+		bumped, err = common.Bump(string(log), common.Minor)
+		if err != nil {
+			return
+		}
+		today := time.Now().Format("02 January 2006")
+		changed = append([]byte(fmt.Sprintf(logUpdate, bumped, today)), log...)
+		return
+	})
+}
+
+const packageFile = "package.json"
+
+// updatePackage updates the package file on the given path with the given version number.
+func updatePackage(path, version string) (err error) {
+	_, err = common.UpdateFile(path, packageFile, func(config []byte) (changed []byte, _ string, err error) {
+		changed, err = common.ReplaceValue(config, "version", version)
+		return
+	})
+	return
+}
+
+const (
+	indexFile     = "index.md"
+	docLinkFormat = `* [v%s%s](http://google.github.io/google-api-nodejs-client/%s/index.html)
+`
+)
+
+// updateIndex updates the documentation index on the given path with the given version number.
+func updateIndex(path, version string) (err error) {
+	_, err = common.UpdateFile(path, indexFile, func(index []byte) (changed []byte, _ string, err error) {
+		changed, _, err = common.ReplacePattern(index, docLinkFormat,
+			fmt.Sprintf(docLinkFormat, version, "$2", version)+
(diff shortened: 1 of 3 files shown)


tcoffee-google:
FYI, I will plan to respond to reviews for this code, but probably attempt no further work on testing or other clients. Based on preliminary inspection, I believe the other libraries will fit into this framework without significant changes.

saicheems:
It's totally fine if you want to leave this to me btw, I'm happy to put it together.

src/main/update/update.go
```diff
@@ -0,0 +1,153 @@
...
+const (
+	// subDirFormat prescribes the subdirectory path pattern for client library subrepos in
+	// discovery-artifact-manager
+	subDirFormat = "clients/%s/google-api-%s-client"
+
+	// repoURLFormat prescribes the remote URL pattern for client library repositories on GitHub
+	repoURLFormat = "https://github.com/google/google-api-%s-client"
+)
+
+func init() {
```
vchudnov-g: Suggest init() at the end of the file

src/main/update/update.go
```diff
@@ -0,0 +1,153 @@
+package update
```
vchudnov-g: Package comment (unless it's already in another file)

src/main/update/update.go
```diff
@@ -0,0 +1,153 @@
...
+// subdirectory, a function regenerating the client library from the local Discovery doc cache, and
+// a function thence returned releasing the updated client library. See: Update.
+var updaters = []updater{
+	{Lib: "nodejs", Update: nodejs.Update},
+	{Lib: "ruby", Update: ruby.Update},
+}
+
+var rootDir string
+
+const (
```
vchudnov-g: Maybe explain the comments what the substitutions are?

src/main/nodejs/nodejs.go
```diff
@@ -0,0 +1,195 @@
...
+		err = fmt.Errorf("Error testing regenerated library: %v", err)
+	}
+
+	release = func() error {
+		if err := common.CheckClean(rootDir); err != nil {
+			return err
+		}
+		err = common.CommandIn(rootDir, "git", "subrepo", "push", subDir).Run()
+		if err != nil {
+			return fmt.Errorf("Error pushing to client library repository (may result from conflicting changes to remote): %v", err)
```
vchudnov-g: nit: I've been reminded that Go error messages should start with lowercase, so if you could lower-case all these.

nit: Also, saying "error" inside the message seems redundant, but I'm not sure what a better alternative is. "Can't push to...." maybe? Saves 3 chars....

src/main/update/update.go
tcoffee-google: Done.

src/main/update/update.go
tcoffee-google: Done.

src/main/update/update.go
tcoffee-google: I think it's clearer here, as it explains up front why the `SubDir` and `RepoURL` fields don't get initialized in the declaration of `updaters`, and what the `...Format` patterns are for.

src/main/nodejs/nodejs.go
tcoffee-google: https://github.com/golang/go/wiki/CodeReviewComments#error-strings gives the rationale that error strings "are usually printed following other context", but my convention actually behaves better in this respect. Instead of
```
error doing task: error executing function foo: error reading file
```
I print
```
Error doing task: Error executing function foo: Error reading file
```
Starting each level with "Error" is specifically intended to aid readability for a deep error trace.

tcoffee-google:
Due to branch cleanup, moving this PR to https://github.com/googleapis/discovery-artifact-manager/pull/37.

src/main/nodejs/nodejs.go
vchudnov-g: I subscribe to deep error traces, but since the outermost error is printed by a `log.Printf`, say, I usually go for messages that wind up more like

"Can't perform task: could not execute foo: could not read file: file 'file bar.txt' not found"

I still maintain that the capitalization and repetition of "error" at each level is non-standard and not helpful (the colons separate the levels of errors)

tcoffee-google: It seems like we're talking about replacing intermediate messages "Error <verb>ing <object>" with "could not <verb> <object>". I fail to see the advantage; but since the mapping is simple, I suggest we let @saicheems establish whatever convention suits him as the update code evolves.

saicheems: Yup, don't worry too much about the details at this point, I haven't yet gotten around to thinking about this system in earnest.

PR #37, opened 2017-04-17
WIP: Automate client library updates

tcoffee-google:
Currently includes Node.js and Ruby. Remains to be tested.

### DIFF
--- a/src/main/common/common.go
+++ b/src/main/common/common.go
@@ -37,6 +37,16 @@ func CheckClean(rootDir string) error {
 	return nil
 }
 
+// PullSubrepo pulls external changes for a subrepository in the given subdirectory of the
+// repository at the given root directory, using the git-subrepo tool. It should not be run
+// concurrently with other operations modifying files in the repository.
+func PullSubrepo(rootDir, subDir string) error {
+	if err := common.CommandIn(rootDir, "git", "subrepo", "pull", subDir).Run(); err != nil {
+		return fmt.Errorf("Error pulling upstream library: %v", err)
+	}
+	return nil
+}
+
 // MaxInt gives the maximum value of the machine-dependent default integer type. (Standard library
 // constants are specific to machine-independent types.)
 const MaxInt = int(^uint(0) >> 1)
@@ -81,7 +91,7 @@ func Bump(versioned string, component int) (bumped string, err error) {
 // UpdateFile rewrites the named file in the given directory by applying the given update function
 // to its contents, returning the modified contents along with any auxiliary information returned by
 // the update function.
-func UpdateFile(dir, name string, update func([]byte) ([]byte, interface{}, error)) (info interface{}, err error) {
+func UpdateFile(dir, name string, update func([]byte) ([]byte, string, error)) (info string, err error) {
 	pathname := path.Join(dir, name)
 	stat, err := os.Stat(pathname)
 	if err != nil {
@@ -103,6 +113,27 @@ func UpdateFile(dir, name string, update func([]byte) ([]byte, interface{}, erro
 		err = fmt.Errorf("Error writing file %s: %v", pathname, err)
 		return
 	}
+	return
+}
+
+// ReplacePattern replaces the first instance, in the input, of a pattern corresponding to a format
+// string by the given change string. It returns a non-nil error if no match appears, otherwise
+// returns the modified input and the modified portion inserted by expanding any template variables
+// in the change string (see: https://golang.org/pkg/regexp/).
+//
+// The format string is assumed to contain substitutions denoted by `%s`. The corresponding regexp
+// pattern is derived by quoting regexp metacharacters and matching substitutions to shortest substrings without newlines
+func ReplacePattern(in []byte, format, change string) (out []byte, inserted string, err error) {
+	var pattern = regexp.MustCompile(strings.Replace(regexp.QuoteMeta(format), "%s", "(.*?)", -1) + `([\s\S]*)`)
+	match := pattern.FindSubmatchIndex(in)
+	if match == nil {
+		err = fmt.Errorf("No match found for pattern `%s`", format)
+		return
+	}
+	insert := string(pattern.ExpandString(nil, format, in, match))
+	left, right := match[0], match[1]
+	out = append(in[:left], append(insert, in[right:]...)...)
+	inserted = string(insert)
 	return
 }
 
--- a/src/main/common/common.go
+++ b/src/main/common/common.go
@@ -15,17 +15,16 @@ import (
 	"strings"
 )
 
-// CommandIn returns the exec.Cmd struct to execute the named program with the given arguments in
-// the specified working directory.
-func CommandIn(dir, name string, arg ...string) *exec.Cmd {
-	cmd := exec.Command(name, arg...)
-	cmd.Dir = dir
+// CommandIn returns the `exec.Cmd` struct to execute `program` with `arguments` in `workingDirectory`.
+func CommandIn(workingDirectory, program string, arguments ...string) *exec.Cmd {
+	cmd := exec.Command(program, arguments...)
+	cmd.Dir = workingDirectory
 	return cmd
 }
 
-// CheckClean verifies that the given repository working directory contains no uncommitted changes.
-func CheckClean(rootDir string) error {
-	diff, err := CommandIn(rootDir, "git", "diff-index", "--quiet", "HEAD").Output()
+// CheckClean verifies that the given repository `rootDirectory` contains no uncommitted changes.
+func CheckClean(rootDirectory string) error {
+	diff, err := CommandIn(rootDirectory, "git", "diff-index", "--quiet", "HEAD").Output()
 	if err != nil {
 		return fmt.Errorf("Error verifying local repository is clean: %v", err)
 	}
@@ -35,11 +34,11 @@ func CheckClean(rootDir string) error {
 	return nil
 }
 
-// PullSubrepo pulls external changes for the given subrepository subdirectory in the given
-// repository root directory, using the git-subrepo tool. It should not be run concurrently with
-// other operations modifying files in the repository.
-func PullSubrepo(rootDir, subDir string) error {
-	if err := CommandIn(rootDir, "git", "subrepo", "pull", subDir).Run(); err != nil {
+// PullSubrepo pulls external changes for the subrepository `subDirectory` in the repository
+// `rootDirectory`, using the git-subrepo tool. It should not be run concurrently with other
+// operations modifying files in the repository.
+func PullSubrepo(rootDirectory, subDirectory string) error {
+	if err := CommandIn(rootDirectory, "git", "subrepo", "pull", subDirectory).Run(); err != nil {
 		return fmt.Errorf("Error pulling upstream library: %v", err)
 	}
 	return nil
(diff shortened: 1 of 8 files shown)


src/main/common/common.go
```diff
@@ -8,11 +8,43 @@ import (
...
+	if err != nil {
+		return fmt.Errorf("Error verifying local repository is clean: %v", err)
+	}
+	if len(diff) != 0 {
+		return errors.New("Local repository contains uncommitted changes")
+	}
+	return nil
+}
+
+// PullSubrepo pulls external changes for a subrepository in the given subdirectory of the
```
vchudnov-g: nit: You could be more concise: "in the subdirectory subDir of the repository at rootDir"

src/main/common/common.go
```diff
@@ -8,11 +8,43 @@ import (
...
+// repository at the given root directory, using the git-subrepo tool. It should not be run
+// concurrently with other operations modifying files in the repository.
```
vchudnov-g: Would it make sense to have a mutex so that at least code within a single binary does not attempt multiple changes concurrently?

src/main/common/common.go
```diff
@@ -54,30 +86,52 @@ func Bump(versioned string, component int) (bumped string, err error) {
 	return
 }
 
-// UpdateFile rewrites the named file by applying the given update function to its contents,
-// returning the modified contents along with any auxiliary information returned by the update
-// function.
-func UpdateFile(name string, update func([]byte) ([]byte, interface{}, error)) (info interface{}, err error) {
-	stat, err := os.Stat(name)
+// UpdateFile rewrites the named file in the given directory by applying the given update function
```
vchudnov-g: nit: Consider s/name/filename/ and "...rewrites filename in directory dir by...' for concision.

src/main/common/common.go
```diff
@@ -54,30 +86,52 @@ func Bump(versioned string, component int) (bumped string, err error) {
...
+// to its contents, returning the modified contents along with any auxiliary information returned by
+// the update function.
+func UpdateFile(dir, name string, update func([]byte) ([]byte, string, error)) (info string, err error) {
+	pathname := path.Join(dir, name)
+	stat, err := os.Stat(pathname)
 	if err != nil {
-		err = fmt.Errorf("Error finding file %s: %v", name, err)
+		err = fmt.Errorf("Error finding file %s: %v", pathname, err)
```
vchudnov-g: convention is error messages are lower-cased

src/main/common/common.go
```diff
@@ -54,30 +86,52 @@ func Bump(versioned string, component int) (bumped string, err error) {
...
+	err = ioutil.WriteFile(pathname, changed, stat.Mode())
 	if err != nil {
-		err = fmt.Errorf("Error writing file %s: %v", name, err)
+		err = fmt.Errorf("Error writing file %s: %v", pathname, err)
+		return
+	}
+	return
+}
+
+// ReplacePattern replaces the first instance, in the input, of a pattern corresponding to a format
```
vchudnov-g: nit: "...the first instance in the input 'in' of a..."

src/main/common/common.go
```diff
@@ -54,30 +86,52 @@ func Bump(versioned string, component int) (bumped string, err error) {
...
+// string by the given change string. It returns a non-nil error if no match appears, otherwise
```
vchudnov-g: Run-on sentence.

src/main/common/updatedisco.go
```diff
@@ -26,23 +26,33 @@ type apiIndex struct {
...
 	if err != nil {
-		return err
+		return
 	}
 
 	updated, errs := writeDiscoCache(index, discoPath, discoDir)
 
 	cleanDiscoCache(discoPath, discoFiles, updated, &errs)
+	err = errs.Error()
+	if err != nil {
```
vchudnov-g: combine into `if errs.Error() != nil`

src/main/ruby/ruby.go
```diff
@@ -0,0 +1,166 @@
...
+
+const (
+	masterBranch = "master"
+
+	// discovery-artifact-manager branch `ruby_release` used for release commits for
+	// google-api-ruby-client
+	releaseBranch = "ruby_release"
+)
+
+func Update(discos []string, rootDir, subDir, repoURL string) (release func() error, err error) {
```
vchudnov-g: I would have a brief function comment saying that this is an `update.update` function.

Also, this solution is fine, but just wanted to make sure you thought about the alternative of defining an `update.Updater` interface that contains a `func release()`, so that this `Update()` here could just return an `update.Updater` rather than a closure. More than one way to approach this, not sure which is simpler. :)

src/main/common/common.go
tcoffee-google: Sorry, I didn't copy over comments from previous PR thread ... here was my take on this: https://github.com/googleapis/discovery-artifact-manager/pull/36#discussion_r111789735

src/main/common/common.go
tcoffee-google: Changed argument name for clarity.

src/main/common/common.go
tcoffee-google: ? ... Added comma for clarity.

src/main/common/updatedisco.go
tcoffee-google: The assignment is needed for the named return.

src/main/ruby/ruby.go
tcoffee-google: > comment

Done.

> alternative

If we wanted to pass the `update.updater` for each language into the `<lang>.Update` function for each language, we'd have to move `update.updater` into `common` to avoid a circular dependency, which seems like a needless complication. I actually prefer the explicit return; the release function will have to be a closure anyway, since it captures variables produced inside the `<lang>.Update` call.

src/main/common/common.go
vchudnov-g: What I mean is that for code that is within the same binary, you could use a mutex to prevent concurrent changes. We can't do anything about changes coming from different binaries, but we can at least prevent this part of the problem.

src/main/common/common.go
vchudnov-g: It's just that so many repetitions of "given" strike me as muddying rather than clarifying. "...rewrites the file in 'name' in directory 'dir' by applying the 'update' function" seems more succinct to me.

src/main/common/common.go
vchudnov-g: Thanks.

nit^2:I think it reads better without the appositive: I would take out the commas around "in the input"

src/main/common/common.go
vchudnov-g: "..if no match appears. Otherwise, it returns ..."
Notice the period (could be a semicolon) and also the "it" subject

src/main/common/updatedisco.go
vchudnov-g: Ah, missed that.

BTW, [here](https://tour.golang.org/basics/7) it says "Naked return statements should be used only in short functions, as with the example shown here. They can harm readability in longer functions." Doing this review yesterday and today, I kind of have to agree. When I read a return, I wonder why you're not returning anything, then I look at the function signature, see the named returns, and then have to hunt for where the variables were last defined.

I would suggest decreasing the use of naked returns. You can still keep the named return values where they help readability.

src/main/ruby/ruby.go
vchudnov-g: OK, just wanted to make sure it was considered.

(But I think in the alternative scenario the closure is not necessary, since you could save state in the struct that implements the proposed interface.)

src/main/common/common.go
tcoffee-google: I've redone all the doc strings to the style you suggest, which works well provided variable names are not too compressed. Seems preferable to me in all respects, if we can get away with the latter.

src/main/common/common.go
tcoffee-google: Done.

src/main/common/common.go
tcoffee-google: Ah, thought you meant the previous sentence. I'm happy to settle a controversial comma splice with a semicolon :-)

src/main/common/updatedisco.go
tcoffee-google: And I always thought that was the best thing about MATLAB :-)  

The pattern (at least in this PR) is always the same: if an error happens, it returns the error and all the other returns are ignored (usually left nil); otherwise, it returns everything with a nil error. Using explicit returns wouldn't save any hunting: even if the return statement includes the identifier, you still have to find where it was actually assigned. It just means that instead of having informative names for all assignments, you might have to parse the position of an expression in a return statement, which surely doesn't help readability.

Seems to me like the big long functions get the most benefit. It sure seems annoying to write ```return nil, nil, nil, fmt.Errorf(...)``` for each of 20 error clauses.

src/main/ruby/ruby.go
tcoffee-google: (Trouble is that the captured state is not necessarily the same for each client language, so using a static struct for this could get pretty ugly.)

src/main/common/common.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// MaxInt gives the maximum value of the machine-dependent default integer type. (Standard library
+// constants are specific to machine-independent types.)
+const MaxInt = int(^uint(0) >> 1)
+
+// Major, Minor, and Patch define group indices of corresponding version number components in
+// versionNumber. Indices are one-based to match subgroup indices in regexp capture.
+const (
+	_ = iota
```
vchudnov-g: nit: I think of `iota` as being useful when you don't particularly care about the specific values of the constants, so new constants can be added or the list rearranged and everything would still work. In this case, you depend on the three fields having specific values, so I would set them individually and explicitly.

src/main/common/common.go
```diff
@@ -0,0 +1,223 @@
...
+
+// ReplacePattern replaces the first instance in `input` of a pattern corresponding to a `format`
+// string, by a `change` string. It returns a non-nil error if no match appears; otherwise, it
+// returns the `modified` input and the `changed` portion obtained by expanding any template
+// variables in the `change` string (see: https://golang.org/pkg/regexp/).
+//
+// The `format` string is assumed to contain substitutions denoted by `%s`. The corresponding regexp
+// pattern is derived by quoting regexp metacharacters and matching substitutions to shortest
+// substrings without newlines.
+func ReplacePattern(input []byte, format, change string) (modified []byte, changed string, err error) {
```
vchudnov-g: Consider a test for this function, though I know we've not been very stringent on having tests for these scripts.

src/main/common/common.go
```diff
@@ -0,0 +1,223 @@
...
+// library `json` package, by the most straightforward means available for our purposes: It uses the
+// standard library decoder to locate the boundaries of the existing value, replacing it with an
+// encoding of the changed value with no extraneous formatting, and leaving the remainder of the
+// object encoding unchanged.
+//
+// Note that Go developers have rejected order-preserving capabilities in the standard library with
+// assertions that JSON is not a human-readable format (the standard library's support for
+// indentation notwithstanding). Compare the second sentence of http://www.json.org, which states
+// that "[JSON] is easy for humans to read and write."
+func ReplaceValue(object []byte, field string, changed interface{}) (modified []byte, err error) {
```
vchudnov-g: Consider a test for this function, though I know we've not been very stringent on having tests for these scripts.

src/main/common/common.go
tcoffee-google: Eliminated named constants, as they seemed to increased verbosity without improving clarity.

src/main/common/common.go
tcoffee-google: Done.

src/main/common/common.go
tcoffee-google: Done.

tcoffee-google:
@saicheems please take a look when you have a chance.

saicheems:
Let's go ahead and merge this, I'll put together a plan to fill in the rest of the languages.

PR #39, opened 2017-04-26
Limit Discovery cache to public APIs

tcoffee-google:
See first commit for code changes.


tcoffee-google:
I've spot checked the file collection. If you have access to an external IP today, you could run the script there for verification.

PR #42, opened 2017-05-05
Update README.md

saicheems:


tcoffee-google:
(I can put my comments in draft form if it's helpful.)

saicheems:
That would be helpful if you have the time - it's hard to get wording right in code review :)

tcoffee-google:
I would word as follows:

# Introduction

The Discovery Artifact Manager is intended to facilitate testing, publishing, and synchronization of generators and artifacts for client libraries and generated code samples of Google APIs defined by the API Discovery Service.

This repo includes copies of the following in separate top-level directories:
- Discovery files from the [API Discovery Service](https://developers.google.com/discovery/)
- the Google API client library generator (used to generate Java and PHP client libraries)
- some of the [Discovery-based Google API client libraries](https://developers.google.com/discovery/libraries), along with their generators
- the code generation [toolkit](https://github.com/googleapis/toolkit/) used to generate code samples for client libraries.

**NOTE**: This repo only contains a cache of the above items; it is not their source of truth. Changes to Toolkit and Discovery-based Google API client libraries should be directed to their respective repos. There is no guarantee that sources or Discovery files in this repo are up to date.

saicheems:
Copy-pasted your recommendations

PR #45, opened 2017-05-19
Add mock and test scripts

saicheems:


saicheems:
This still isn't ready for review - I need to clean up the code, remove some repetitive bits, and write a small doc for how everything fits together...

but it works!! take a look at the travis run, the command `python run_tests.py pubsub:v1 tasks:v1` runs each sample in each language against the mock server. This exercise has revealed quite a few bugs, and some behavior, which while technically isn't a bug, that we didn't know was present in the client libraries.

saicheems:
Again please hold off on review until I give the all green, just wanted to keep you guys up to date on this PR.

saicheems:
I'm going to claim this is ready.

It's all a ton of new code, so I'll try to explain how things tie together, and with luck it won't be outrageous to review.

There are essentially 2 parts:

1. The `mockgen` module.

This Python module contains a number of scripts which are responsible for generating modified discovery documents, override files, and a mock server for testing purposes.

* `generate_mock_discovery_document.py`

Generates a mock Discovery document with all paths which are ambiguous modified to be unambiguous. For example:

If two functions declare their path as `/foo/{bar}` and `/foo/{baz}`, those paths are modified to be unique by prepending a UUID: `/1/foo/{bar}` and `/2/foo/{baz}`.
Note that if two methods have the same path but different HTTP verbs, they aren't modified. Only methods which have the same path and HTTP verb are disambiguated.

The output Discovery document also points to `http://localhost:8000` (the `rootUrl` field is changed).

* `generate_mock_value_override.py`

Generates a mock override file which sets non-trivial values for required string fields with patterns.

For example, it may replace the path template string `foo/my-foo` with `foo/abcd1234` in the sample.

* `generate_mock_server.py`

Generates a mock server which implements each method in the input Discovery document. The expected input is the output of `generate_mock_discovery_document.py`.

2. The `run_tests.py` script.

Accepts a list of discovery documents and runs each sample against a generated mock server. The script initializes the language environment and generates/compiles all samples before execution. Check the Travis build for an example of the output.

run_tests.py
```diff
@@ -0,0 +1,744 @@
...
+
+def _make_lib_google_api_client_generator(test_dir):
+    lib_dir = _make_lib_dir(test_dir)
+    client_generator_dir = os.path.join(lib_dir, 'google-api-client-generator')
+    if not os.path.exists(client_generator_dir):
+        cmd = 'cp -r google-api-client-generator {}'.format(lib_dir)
+        subprocess.call(shlex.split(cmd))
+        # TODO: Note that we use Python2 b/c 3 doesn't work.
+        cmd = 'virtualenv venv'
+        subprocess.call(shlex.split(cmd), cwd=client_generator_dir)
```
vchudnov-g: It would make the file more readable if you could  modify these two-line subprocess calls into a single logical line that reduced boilerplate as much as possible. For example

`run('blah blah {}foo bar'.format(...),
      cwd=my_dir, env=my_env)
`

run_tests.py
```diff
@@ -0,0 +1,744 @@
...
+def _make_src_dir(test_dir, name, version, language=None):
+    src_dir = os.path.join(test_dir, 'src', name, version)
+    if language:
+        src_dir = os.path.join(src_dir, language)
+    if not os.path.exists(src_dir):
+        os.makedirs(src_dir)
+    return src_dir
+
+
+def _generate_overrides(test_dir, discovery_document_filename, name, version):
```
vchudnov-g: Could you provide comments for at least the functions that are not trivial/obvious, and for code stanzas that are not obvious?

saicheems: Ah, yes, I forgot to run through and comment `run_tests.py`, will do so today.

saicheems: Added docs to most of the functions, and peppered some line comments here and there...

I'll try to make another pass to add more comments in some time, please let me know if there's anything in particular that you want more exposition for.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+            return self._gen_type(param, visited)
+        type_ = schema['type']
+        if type_ == 'any':
+            return {}
+        if type_ == 'array':
+            return [self._gen_type(schema['items'], visited)]
+        if type_ == 'boolean':
+            return False
+        if type_ == 'integer':
+            return {'int32': 42, 'uint32': 42}.get(schema.get('format', 42))
```
tcoffee-google: Did you mean
```python
{'int32': 42, 'uint32': 42}.get(schema.get('format'), 42)
```
?

tcoffee-google: Should we give (and test) values that distinguish these integer formats from one another---for instance, to check against any client library action that does so?

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        if type_ == 'number':
+            return {'double': 42, 'float': 42}.get(schema.get('format', 42))
```
tcoffee-google: Similar comments to above.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        if type_ == 'object':
+            obj = {}
+            id_ = schema.get('id')
+            if id_ in visited:
```
tcoffee-google: Comment that nested objects don't have IDs is also relevant here. Maybe this and the logic for `visited.add` could be combined together to simplify?

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+                visited.add(id_)
+            for key, val in six.iteritems(schema.get('properties', {})):
+                obj[key] = self._gen_type(val, visited)
+            return obj
+        if type_ == 'string':
+            return {
+                'byte': 'foo',
+                'date': '1970-01-01',
+                'date-time': '1970-01-01T00:00:00-07:00',
+                'int64': '42',
```
tcoffee-google: Again, should we test values distinguishing the two integer formats?

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,104 @@
...
+import rstr
+
+import discoveryutil
+import six
+
+
+def _gen_fields(method, quote='"'):
+    """Returns the override dict for required values for the given method.
+
+    Required fields are assigned a non-zero value that should be accepted by
```
tcoffee-google: More precise specification than "non-zero"?

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,104 @@
...
+    all client libraries.
+
+    Args:
+        method (dict): The Discovery method.
+        quote (string, optional): The symbol to enquote strings with.
```
tcoffee-google: Clearer to call something like "delimiter for string literals".

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,104 @@
...
+
+    Returns:
+        dict: An override mapping for fields in the given method which must
+            have a non-trivial value.
+    """
+    params = method.get('parameters', {})
+
+    fields = {}
+    for name, param in six.iteritems(params):
+        if not param.get('required'):
```
tcoffee-google: I seem to recall that this condition has not always been reliable across Discovery documents, whereas `parameterOrder` has been --- can you either use the latter or cross-check these against each other?

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,104 @@
...
+        pattern = param.get('pattern')
+        if not pattern:
+            is_string = param.get('type') == 'string'
+            has_format = bool(param.get('format'))
+            is_repeated = param.get('repeated')
+
+            # TODO: Add support for required repeated fields.
+
+            # At the moment it's only useful to override strings which have no
+            # format specified (don't bother with bytes or dates). Required
```
tcoffee-google: Is this because our samples already provide non-trivial values for the others? Might clarify.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+                'uint64': '42',
+            }.get(schema.get('format'), 'foo')
+        raise Exception('unexpected type: {}'.format(type_))
```
tcoffee-google: Do we need support for enums here?

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,104 @@
...
+import json
+import random
+
+import rstr
+
+import discoveryutil
+import six
+
+
+def _gen_fields(method, quote='"'):
```
tcoffee-google: It looks like this function currently only handles strings, so I must be confused. Why not share logic with `_gen_types` for this purpose?

run_tests.py
```diff
@@ -0,0 +1,940 @@
...
+        test_dir (string): The parent directory.
+
+    Returns:
+        string: The path to the lib/google-api-client-generator directory.
+    """
+    lib_dir = _make_lib_dir(test_dir)
+    client_generator_dir = os.path.join(lib_dir, 'google-api-client-generator')
+    if not os.path.exists(client_generator_dir):
+        _call('cp -r google-api-client-generator {}'.format(lib_dir))
+        # TODO: Note that we use Python2 b/c 3 doesn't work.
```
tcoffee-google: Let's remember to track in an issue.

run_tests.py
```diff
@@ -0,0 +1,940 @@
...
+
+    # 3. Generate the auth/discoveryDocUrl override.
+    misc_override_filename = 'override{}.json'.format(suffix)
+    misc_override_filename = os.path.join(src_dir, misc_override_filename)
+    filenames.append(misc_override_filename)
+    with open(misc_override_filename, 'w') as file_:
+        url = 'http://localhost:8000/static/{}.{}.json'
+        url = url.format(name, version)
+        override = {
+            'csharp|go|java|js|nodejs|php|python|ruby': {
```
tcoffee-google: I forget, can we just use `'*'` for this still?

run_tests.py
```diff
@@ -0,0 +1,940 @@
...
+
+def _generate_samples(ctx, language, ruby_names_file=None):
+    """Generates and returns the filenames of samples for an API/language.
+
+    All samples are written to a temporary directory.
+
+    Args:
+        ctx (Context): The Context to generate from.
+        language (string): The language to generate for.
+        ruby_names_file (string, optional): The path to the Ruby names file.
```
tcoffee-google: Can the update to this file also be automated?

run_tests.py
```diff
@@ -0,0 +1,940 @@
...
+    Returns:
+        list: A list of Commands to run samples.
+    """
+    lib_dir = _make_lib_dir(test_dir)
+    client_lib_dir = os.path.join(lib_dir, 'google-api-dotnet-client')
+
+    # Clone the client and reset it to v1.26.2 (latest).
+    _call('git clone https://github.com/google/google-api-dotnet-client',
+          cwd=lib_dir)
+    _call('git reset --hard v1.26.2', cwd=client_lib_dir)
```
tcoffee-google: What's the plan to deal with client library updates?

run_tests.py
```diff
@@ -0,0 +1,940 @@
...
+    Args:
+        ctx (Context): The Context to generate from.
+        language (string): The language to generate for.
+        ruby_names_file (string, optional): The path to the Ruby names file.
+
+    Returns:
+        list: A list of generated sample filenames.
+    """
+    temp_dir = tempfile.mkdtemp()
+    cmd = ('java -jar discoGen-0.0.5.jar'
```
tcoffee-google: What's the plan to deal with `discoGen` updates?

mockgen/discoveryutil/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        string: A path signature.
+    """
+    path = method.get('flatPath', method['path']).strip()
+    path = re.sub(r'{[\+][^}]*}', '{+}', path)
+    path = re.sub(r'{[^\+][^}]*}', '{}', path)
+    path = '{}:{}'.format(path, method['httpMethod'])
+    return path
+
+
+def parse_methods(root, methods=None):
```
saicheems: The specific structure that it allows is:

```
{
  ...
  "methods": { ... }
  "resources": {
    "resource1": { "methods": { ... } },
    "resource2": { "resources": { "resource3": { "methods": { ... } } } }
}
...
```

The mechanism you suggested is missing the recursive step for exploring sub-resources

mockgen/mockgen/generate_mock_discovery_document.py
```diff
@@ -0,0 +1,74 @@
...
+from __future__ import absolute_import
+import argparse
+import hashlib
+import json
+
+import discoveryutil
+import six
+
+
+def _disambiguate_method_paths(root, seen=None):
```
saicheems: https://github.com/googleapis/discovery-artifact-manager/pull/45#discussion_r122053289

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+
+
+class _Generator(object):
+    """A Generator which emits a mock server from a Discovery document."""
+
+    _CAST_FUNC = {
+        'any': 'dict',
+        'array': 'list',
+        'boolean': 'bool',
+        'integer': 'int',
```
saicheems: This script runs w/ Python3, where integers no longer have limits

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        w('')
+        # ApiError represents a "Bad Request" exeception raised by the mock
+        # server if parameters of the incomding request is determined to be
+        # invalid.
+        w('class ApiError(Exception):')
+        w('    def __init__(self, msg, code=400):')
+        w('        self.message = msg')
+        w('        self.code = code')
+        w('')
+        w('    def to_dict(self):')
```
saicheems: I went to go look, but I couldn't find any official docs :/

This format was derived from the responses that are returned from API Explorer and the source of google-api-ruby-client:

https://github.com/google/google-api-ruby-client/blob/master/lib/google/apis/core/api_command.rb#L152

I believe it's the original error format, so the proto is around somewhere internally.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+
+        # The only way that 'type' is not a property of param is if it contains
+        # a reference to another schema. In that case we assume it's an object.
+        type_ = param.get('type', 'object')
+        cast_func = self._CAST_FUNC[type_]
+        instance = self._INSTANCE[type_]
+
+        location = param['location']
+        # The Ruby client library may send query parameters as part of the form
+        # if possible.
```
saicheems: It's not, just a consequence of the many ways to send a valid HTTP request. Technically, Ruby's choice is valid, but so are the choices of every other client to not embed data into the form :/

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        w('    request_args = request.args or request.form')
+        if location == 'query':
+            w('    if "{}" not in request_args:'.format(name))
+            msg = 'query parameter \\"{}\\" not found'.format(name)
+            w('        raise ApiError("{}")'.format(msg))
+            w('    try:')
+            cast_func = self._CAST_FUNC[type_]
```
saicheems: Not necessarily, because query args come in as strings. Plus, end up come in as strings.

mockgen/mockgen/generate_mock_server.py
saicheems: No, because enums are actually string types. They just happen to have a set of values they support, but none of the client libraries actually validate that set.

run_tests.py
saicheems: It could... but the way this program is organized makes it weird to do so - the client library for ruby would have to be generated before the toolkit jar is generated. I decided it was _nicer_ for the structure of the program to pass a flag instead.

run_tests.py
saicheems: The jar was admittedly a placeholder because I didn't want to wait for toolkit to build on Travis, I'll change it to generate from toolkit before this is merged.

Though if we keep just the jar, we don't have to have the toolkit directory anymore. In that either case, the story for updates is the same. Toolkit has to be pulled periodically, or discoGen has to be replaced with a more recent jar periodically.

Now that I mention it though, would you be satisfied with just keeping the jar file? It shaves some ~10 minutes of the Travis builds.

run_tests.py
saicheems: I fixed it to a version because dotnet master was in a broken state for a while... change it to work on master directly, which works now, or to periodically update the revision to fix to in this script.

mockgen/mockgen/generate_mock_value_override.py
saicheems: I did earlier, when I was writing the new Discovery parser in Toolkit. I found that `required` was reliable in all APIs except one, which is Google internal and *very* old.

I don't think a failure is possible anymore, and I do think that we can trust the `required` field like we do for `parameterOrder`.

mockgen/mockgen/generate_mock_value_override.py
saicheems: Done

mockgen/mockgen/generate_mock_value_override.py
saicheems: Added some qualification.

mockgen/mockgen/generate_mock_value_override.py
saicheems: Theoretically we could share some logic, but I think the result would be more difficult to understand than just being explicit.

Plus, `_gen_fields` really has more to do with generating string values in an overrides file, which is subtly different than `_gen_types`, which generates non-trivial objects from `Schema` definitions.

mockgen/mockgen/generate_mock_server.py
saicheems: Done

mockgen/mockgen/generate_mock_server.py
saicheems: Done

mockgen/mockgen/generate_mock_server.py
saicheems: Yes, good catch, thanks

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+            w('        {}(request_args.get("{}"))'.format(cast_func, name))
+            w('    except:')
+            msg = 'expected \\"{}\\" to be an instance of \\"{}\\"'
+            msg = msg.format(name, instance)
+            w('        raise ApiError("{}")'.format(msg))
+        else:  # location == 'path'
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+            param (dict): A Discovery schema.
+        """
+        w = _w(file_)
+
+        # TODO: Because the samples are generated with array fields initialized
+        # as empty arrays, client libraries may interpret the value as null. As
+        # a result, they may not send anything for the field, which results in
+        # a 400 error.
+        # This has to be fixed in the samples by initializing each array with a
+        # single entry of its type (ex: [''] instead of []).
```
saicheems: https://github.com/googleapis/toolkit/issues/1377

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        # Emit the response.
+        if 'response' in method:
+            ref = method['response']['$ref']
+            obj = self._gen_type(self._schemas[ref])
+            # If "dataWrapper" is one of the values in the top-level "features"
+            # key, then the client library expects the response to be nested
+            # under the key "data".
+            if 'dataWrapper' in self._features:
+                obj = {'data': obj}
+            # We pop the page token key from the response object because page
```
saicheems: https://github.com/googleapis/discovery-artifact-manager/issues/50

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        w('@app.route("/{}", methods={})'.format(path, json.dumps(http_verbs)))
+        method_name = method['id'].replace('.', '_')
+        w('def {}({}):'.format(method_name, ', '.join(url_vars)))
+
+        params = method.get('parameters', {})
+        # "parameterOrder" contains a list of required parameters.
+        param_order = method.get('parameterOrder', {})
+        for name in param_order:
+            location = params[name]['location']
+            # If name is a path variable and not in url_vars then it represents
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        # be multi-segment variables (prepended with a "+", ex: "{+foo}") or
+        # single-segment variables (ex: "{foo}"). We pull all variables out of
+        # the path and escape them with an underscore to prevent naming
+        # conflicts.
+        # ex: "/foo/{bar}/{+baz}" => ["bar_", "baz_"]
+        url_vars = [_esc_var(x) for x in re.findall(r'{\+?([^}]+)}', path)]
+
+        # Convert path into a Flask route.
+        # ex: "{+baz}" => "<path:{}>"
+        path = re.sub(r'{(\+[^}]+)}', '<path:{}>', path)
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        w('        if method:')
+        w('            method = method.encode("ascii", "replace")')
+        w('            environ["REQUEST_METHOD"] = method')
+        w('        return self.app(environ, start_response)')
+        w('')
+        w('app = Flask(__name__)')
+        w('app.wsgi_app = HTTPMethodOverrideMiddleware(app.wsgi_app)')
+        w('')
+        # ApiError represents a "Bad Request" exeception raised by the mock
+        # server if parameters of the incomding request is determined to be
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        # Emit a middleware class to handle the "HTTP-Method-Override" header.
+        # The Java client library sends "PATCH" requests as "POST" requests
+        # with the "HTTP-Method-Override" header set to "PATCH".
+        w('class HTTPMethodOverrideMiddleware(object):')
+        w('    def __init__(self, app):')
+        w('        self.app = app')
+        w('')
+        w('    def __call__(self, environ, start_response):')
+        w('        method = environ.get("HTTP_X_HTTP_METHOD_OVERRIDE", "")')
+        w('        method = method.upper()')
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        w('import gzip')
+        w('import io')
+        w('from flask import Flask')
+        w('from flask import jsonify')
+        w('from flask import request')
+        w('')
+        # Emit a middleware class to handle the "HTTP-Method-Override" header.
+        # The Java client library sends "PATCH" requests as "POST" requests
+        # with the "HTTP-Method-Override" header set to "PATCH".
+        w('class HTTPMethodOverrideMiddleware(object):')
```
saicheems: I'm not sure it's worth the trouble to break apart this class, IMO it's easier to work with it if everything is in one place and flows without module calls.

The suffix `Middleware` is idiomatic for web servers, and Flask applications in particular, so I think it's okay. I didn't know of the word "hifalutin" before though, so TIL hahaha

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+
+        Args:
+            file_ (File): The file to write to.
+        """
+        w = _w(file_)
+
+        # Emit imports and the initialization code for the Flask server.
+        w('import gzip')
+        w('import io')
+        w('from flask import Flask')
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
+"""Generates a mock Discovery service server from a Discovery document.
+
+The generated server is configured to run on "http://localhost:8000" and
+contains an implementation of each method in the given Discovery document.
+For each method, the generated server:
+ - performs asserts to validate input where useful.
+ - returns a non-trivial and complete response.
```
saicheems: Dropped complete, added a bit of qualification:

`- returns a non-trivial (non-zero) response.`

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,466 @@
...
+        cast_func = self._CAST_FUNC[type_]
+        instance = self._INSTANCE[type_]
+
+        location = param['location']
+        # The Ruby client library may send query parameters as part of the form
+        # if possible.
+        w('    request_args = request.args or request.form')
+        if location == 'query':
+            w('    if "{}" not in request_args:'.format(name))
+            msg = 'query parameter \\"{}\\" not found'.format(name)
```
saicheems: I'll try to consolidate as bit. I was trying to avoid ugly line breaking

mockgen/mockgen/generate_mock_server.py
saicheems: Hmmm... do you think maybe using 2^32-1 and 2&64-1 as values instead?

mockgen/mockgen/generate_mock_server.py
saicheems: https://github.com/googleapis/discovery-artifact-manager/pull/45#discussion_r123364214

run_tests.py
saicheems: I think we're actually good on this one now that the `0L` thing is fixed, I'll drop the todo.

run_tests.py
saicheems: Just checked, it doesn't work.

mockgen/mockgen/generate_mock_server.py
saicheems: Changed to use the max values for the data types.

saicheems:
Ready for another look

mockgen/mockgen/generate_mock_server.py
vchudnov-g: I don't know that we need a separate module, but I strongly encourage having each of the two boilerplates (before and after the `_emit_method` loop) be defined as a single multi-line string and written once.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+            file_ (File): The file to write to.
+        """
+        w = _w(file_)
+
+        # Emit imports and the initialization code for the Flask server.
+        w('import gzip')
+        w('import io')
+        w('from flask import Flask, jsonify, request')
+        w('')
+        # Emit a middleware class to handle the "HTTP-Method-Override" header.
```
vchudnov-g: I suggest including the comments about the code in the emitted code (and omitting the references to "emit" in those comments).

In `_emit_method` some of the comments are genuinely clarifying what you're emitting, so those should stay as is and not get passed along.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+    def emit(self, file_):
+        """Emits a mock server.
+
+        Args:
+            file_ (File): The file to write to.
+        """
+        w = _w(file_)
+
+        # Emit imports and the initialization code for the Flask server.
+        w('import gzip')
```
vchudnov-g: I suggest adding a file comment at the very top saying this is an autogenerated server and including metadata such as the time.

mockgen/discoveryutil/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    """Parses methods from the given Discovery document.
+
+    Args:
+        root (dict): A Discovery document.
```
saicheems: Done

mockgen/discoveryutil/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+        methods (dict): A mapping of method ID to Discovery method. Do not set.
```
saicheems: Done

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,115 @@
+"""Generates a mock override file from a Discovery document.
+
+The generated file contains override mappings for the default value of all
+fields which must have a non-trivial value.
```
saicheems: Done

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,115 @@
...
+            # At the moment it's only useful to override strings which have no
+            # format specified (don't bother with bytes or dates). Required
+            # strings in several libraries cannot be empty, so we set their
+            # value to be a single space.
+            if is_string and not has_format and not is_repeated:
+                value = 'foo'
+                if param.get('enum'):
+                    value = param.get('enum')[0]
+                fields[name] = {'defaultValue': quote + value + quote}
+            # TODO: In the Node.js client, required integers must be non-zero.
```
saicheems: Actually it turns out I fixed this bug a little while ago myself:
https://github.com/google/google-api-nodejs-client/pull/757

So I removed the TODO.

mockgen/mockgen/generate_mock_value_override.py
```diff
@@ -0,0 +1,115 @@
...
+
+    # A dict of overrides for languages that use double quotes.
+    double_quote = {}
+    # A dict of overrides for languages that use single quotes.
+    single_quote = {}
+    for id_, method in six.iteritems(methods):
+        fields = _gen_fields(method)
+        if not fields:
+            continue
+        double_quote[id_] = {'fields': fields}
```
saicheems: I tried it, and IMO it's not as obvious as it is right now, so I'd prefer to leave it

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+
+
+_PROXY_HTML = """<!DOCTYPE html>
+<html>
+<head>
+<title></title>
+<meta http-equiv="X-UA-Compatible" content="IE=edge" />
+<script type="text/javascript">
+  window['startup'] = function() {
+    googleapis.server.init();
```
saicheems: Deleted this whole block, the `proxy.html` script is required to run the JS samples (it loads and initializes parts of the JS library through some magic), but I'm not willing to set up Selenium or headless Chrome at this time.

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+  };
+</script>
+<script type="text/javascript"
+  src="https://apis.google.com/js/googleapis.proxy.js?onload=startup" async
```
saicheems: https://github.com/googleapis/discovery-artifact-manager/pull/45#discussion_r124713241

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+
+The JavaScript client library expects this file under
+"{rootUrl}/static/proxy.html".
+"""
+
+
+class _Generator(object):
+    """A Generator which emits a mock server from a Discovery document."""
+
+    _CAST_FUNC = {
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+        'any': 'dict',
+        'array': 'list',
+        'boolean': 'bool',
+        'integer': 'int',
+        'number': 'float',
+        'object': 'dict',
+        'string': 'str'
+    }
+
+    _INSTANCE = {
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
```diff
@@ -0,0 +1,478 @@
...
+        for id_, method in six.iteritems(self._methods):
+            path_signature = discoveryutil.path_signature(method)
+            if path_signature in path_signatures:
+                msg = 'method "{}" and "{}" have the same path'
+                msg = msg.format(path_signatures[path_signature], id_)
+                raise Exception(msg)
+            path_signatures[path_signature] = id_
+
+        self._features = root.get('features', [])
+        schemas = {}
```
saicheems: Done

mockgen/mockgen/generate_mock_server.py
saicheems: Done

mockgen/mockgen/generate_mock_server.py
saicheems: Done. Good call, that made it a lot easier to read. Basically combined all the static text into one long multi-line string.

mockgen/mockgen/generate_mock_server.py
saicheems: Done

PR #46, opened 2017-06-13
Fix 404 and past revision behavior

saicheems:
1. Prevent APIs from being updated with past versions

   Sometimes, multiple consecutive requests to the Discovery service
   return Discovery documents with different revision tags.

   This commit adds a check to verify that the downloaded Discovery
   document has a revision that is at least as recent as the one that
   already exists (if one already exists).

   If the revision check fails, the update process for that particular
   Discovery document is aborted as a no-op without an error.

2. Error on non-200 Discovery document responses

   This commit also adds a check for non-200 responses from Discovery
   URLs. There are some cases where the Discovery directory lists APIs
   which point to missing Discovery documents. In the case that a
   non-200 response is returned from a Discovery document URL, an error
   is returned, but any existing Discovery document for that API version
   is not removed.

### DIFF
--- a/src/main/common/updatedisco.go
+++ b/src/main/common/updatedisco.go
@@ -204,6 +204,14 @@ func UpdateAPI(API apiInfo, absolutePath string, permissions os.FileMode, update
 		return fmt.Errorf("Error parsing Discovery doc from %v: %v", API.DiscoveryRestURL, err)
 	}
 
+	// If "revision" is nil or not a string, the empty string is returned.
+	newRevision, _ := newAPI["revision"].(string)
+	oldRevision, _ := oldAPI["revision"].(string)
+	// Do nothing if the revision of the new API is older than what already exists.
+	if newRevision < oldRevision {
+		return fmt.Errorf("Error validating Discovery doc revision %v: %v < %v", newRevision, oldRevision)
+	}
+
 	if oldAPI == nil || !sameAPI(oldAPI, newAPI) {
 		if err := ioutil.WriteFile(filepath, newDisco, permissions); err != nil {
 			return fmt.Errorf("Error writing Discovery doc to %v: %v", filepath, err)
@@ -236,6 +244,17 @@ func discoFromURL(URL string) (contents []byte, err error) {
 		return
 	}
 	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
+		// Fail if the status code is not 200.
+		// This can happen if a service is listed in the Discovery
+		// directory, but the Discovery document is not accessible.
+		// In this case, the existing Discovery document is left in
+		// place until the directory is updated to delist the service.
+		// At that point, the updatedisco script will delete the
+		// Discovery document.
+		err = fmt.Errorf("Error downloading Discovery doc from %v: %v response", URL, response.StatusCode)
+		return
+	}
 	contents, err = ioutil.ReadAll(response.Body)
 	if err != nil {
 		err = fmt.Errorf("Error reading Discovery doc from %v: %v", URL, err)
--- a/src/main/common/updatedisco.go
+++ b/src/main/common/updatedisco.go
@@ -209,7 +209,7 @@ func UpdateAPI(API apiInfo, absolutePath string, permissions os.FileMode, update
 	oldRevision, _ := oldAPI["revision"].(string)
 	// Do nothing if the revision of the new API is older than what already exists.
 	if newRevision < oldRevision {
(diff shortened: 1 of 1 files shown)


tcoffee-google:
Just to be clear, are you saying that two consecutive calls to Discovery service can return two Discovery documents for which (a) the revision of the second is earlier than that of the first, and (b) the substantive contents (as checked by the existing script) of the second is different from that of the first?

saicheems:
Yes, if you take a look down this list:

https://github.com/saicheems/discovery-artifact-manager/commits/master

You'll see that, for example, the `ml.v1.json` document gets updated nearly every other commit. The change is nontrivial and not caught by equality checks.

It's a bit hard to trigger it by just refreshing a bunch, but here's the URL for that API:
`https://ml.googleapis.com/$discovery/rest?version=v1`

saicheems:
Also, just changed the behavior so it's an error if a 404 is returned. I think it's more useful if that's logged.

saicheems:
https://github.com/saicheems/discovery-artifact-manager/commit/8158a0ee53df3df61d5ca5169c99b1ccd29462f9

^ In this commit, the latest revision for `ml:v1`, `20170612` is replaced with `20170604`

src/main/common/updatedisco.go
```diff
@@ -204,6 +204,14 @@ func UpdateAPI(API apiInfo, absolutePath string, permissions os.FileMode, update
 		return fmt.Errorf("Error parsing Discovery doc from %v: %v", API.DiscoveryRestURL, err)
 	}
 
+	// If "revision" is nil or not a string, the empty string is returned.
+	newRevision, _ := newAPI["revision"].(string)
+	oldRevision, _ := oldAPI["revision"].(string)
```
tcoffee-google: Check for `oldAPI == nil`?

src/main/common/updatedisco.go
```diff
@@ -236,6 +244,17 @@ func discoFromURL(URL string) (contents []byte, err error) {
 		return
 	}
 	defer response.Body.Close()
+	if response.StatusCode != http.StatusOK {
```
tcoffee-google: So the above error clause will not trigger in the case of 404? If so, please mention in the comment to clarify why this check is not redundant.

saicheems:
Changed it one more time - both cases are returned as errors now. I think it's better for logging purposes.

tcoffee-google:
> the ml.v1.json document gets updated nearly every other commit

These revisions are more than a week apart ... should we file a bug against Discovery?

src/main/common/updatedisco.go
saicheems: An empty map is returned from parseAPI, so that check is not required - if `oldAPI` is an empty map, `oldRevision` will be the empty string.

src/main/common/updatedisco.go
saicheems: Are you talking about the error clause for the `Get` call? I feel that's straightforward, a failure in that case is a transport or network issue - a 404 is not a failure for `http.Get`.

Do you want me to comment that a non-200 response doesn't return an error?

saicheems:
I'm not sure - technically, the old revision was correct as of 1~2 days ago, so it might resolve itself over the next day as last weeks version is updated in some cache.

src/main/common/updatedisco.go
tcoffee-google: OK, the comment is sufficient.

src/main/common/updatedisco.go
tcoffee-google: Seems helpful to me, since it otherwise looks odd that there are two different error clauses for "error downloading"; or perhaps even clearer:
```go
 	response, err := http.Get(URL)
	defer response.Body.Close()
	if err == nil && response.StatusCode != http.StatusOK {
 		err = fmt.Errorf("HTTP error: %v", response.Status)
 	}
 	if err != nil {
 		err = fmt.Errorf("Error downloading Discovery doc from %v: %v", URL, err)
  		return
	}
```
(I think this also fixes unclosed response bodies in case of error.)

saicheems: Added a comment to clarify - I personally prefer the existing structure, so I left that in place. I feel that checking for `err == nil` is more confusing.

I'm also not sure if it's the right practice to make the defer call before checking that `err != nil`... the documentation is somewhat ambiguous, but suggests that response could be `nil` if err is not `nil`: `"When err is nil, resp always contains a non-nil resp.Body. Caller should close resp.Body when done reading from it."`

PR #47, opened 2017-06-14
Rename "Web" to "Browser"

saicheems:


tcoffee-google:
+1

saicheems:
My only hesitation against "JavaScript (Browser)" is that the tab name is too long. The next longest tab name would be "Python". I think this would look weird, but not sure how you feel about it.

tcoffee-google:
My position has always been that "JavaScript" is the best choice. I've never heard anyone say that they program in Browser.

saicheems:
I realize there's little chance of 100% agreement, but given that there's internal consensus on "Browser", can we continue with it? Considering the original naming was "Web"... it is an improvement.

vchudnov-g:
It's your call, @saicheems 

From where I stand, the number one priority by far is consistency across pages on our doc sites. After that, I would say it's important (but less so) to have consistency in the tab naming, which is why I prefer "JavaScript (Browser)" from among the options favored by the team who asked for this. But you're right that "Browser" is no worse than "Web" here.

With either of these two options, this pr lgtm.

tcoffee-google:
I concur with all the above.

saicheems:
Will roll with this, and come back if there's further complaints. We're only starting with one API now anyway, so I hope it won't be a big issue.

vchudnov-g:
Sure, but we should republish the samples for the rolled-out APIs for consistency.

saicheems:
We haven't rolled any out with JS have we?

saicheems:
(since it's disabled by default, no APIs show JS samples right now)

vchudnov-g:
Oh, I think you're right. We never did manage to publish anything with JS earlier. Nm.

PR #49, opened 2017-06-16
Don't clean out index.json every other run...

saicheems:


saicheems:
Overlooked this case in the previous PR, the clean function deletes `index.json` because it isn't an API - so every other call replaces the file entirely.

Note that I don't think we need to do deep equality checks for the index because as far as I can tell it's always returned in a deterministic order.

saicheems:
Moved to `readDiscoCache`

src/main/common/updatedisco.go
```diff
@@ -73,6 +74,13 @@ func readDiscoCache() (absolutePath string, directory os.FileInfo, files []os.Fi
 	if err != nil {
 		err = fmt.Errorf("Error reading path for Discovery docs: %v", absolutePath)
 	}
+	// Remove "index.json" from files, as it's not a Discovery document.
+	for i := 0; i < len(files); i += 1 {
+		if files[i].Name() == "index.json" {
+			files = append(files[:i], files[i+1:]...)
```
tcoffee-google: While I can't strictly verify from the documentation, I'm assuming the underlying representation of `FileInfo` is deeply pointer-free, so this does not leak memory. You could optimize as
```golang
files[i] = files[len(files)-1]
files = files[:len(files)-1]
```
However, it looks to me like we don't actually need `FileInfo` objects here, just file names, so you could use [`filepath.Glob`](https://golang.org/pkg/path/filepath/#Glob) instead.

vchudnov-g: Using `filepath.Glob` is not a bad idea.

I'm inclined to vote against the list optimization because it's less clear and I don't think it would buy us much. In fact, I don't think this achieves anything in Go: If we modify the existing slice to have one fewer element, the underlying array is still being referenced and thus no parts of it are released (https://blog.golang.org/go-slices-usage-and-internals), whereas appending to a new slice will create a new, shorter array and the old one will shortly be gc'd (https://blog.golang.org/slices)

tcoffee-google: I meant optimization for time, not space, but no big deal.

saicheems: Switched to `Glob`

src/main/common/updatedisco.go
```diff
@@ -41,23 +42,24 @@ func UpdateDiscos() (names []string, err error) {
...
+		path := path.Join(absolutePath, filename)
+		names = append(names, path)
 	}
 	return
 }
 
 // readDiscoCache returns the `absolutePath` to the top-level 'discoveries' directory along with its
-// `directory` attributes and those of all discovery `files` therein.
-func readDiscoCache() (absolutePath string, directory os.FileInfo, files []os.FileInfo, err error) {
+// `directory` attributes and those of all discovery `filepaths` therein. Note that the discovery
```
tcoffee-google: s/those of //
(now returning paths to files rather than attributes)

src/main/common/updatedisco.go
```diff
@@ -69,9 +71,19 @@ func readDiscoCache() (absolutePath string, directory os.FileInfo, files []os.Fi
 		err = fmt.Errorf("Error finding path for Discovery docs: %v", absolutePath)
 		return
 	}
-	files, err = ioutil.ReadDir(absolutePath)
+	globPath := path.Join(absolutePath, "*.json")
+	filepaths, err = filepath.Glob(path.Join(absolutePath, "*.json"))
```
tcoffee-google: Use `globPath` here?

src/main/common/updatedisco.go
```diff
@@ -69,9 +71,19 @@ func readDiscoCache() (absolutePath string, directory os.FileInfo, files []os.Fi
...
 	if err != nil {
-		err = fmt.Errorf("Error reading path for Discovery docs: %v", absolutePath)
+		err = fmt.Errorf("Error globbing path for Discovery docs: %v", globPath)
```
tcoffee-google: Would prefer s/globbing/searching/.

src/main/common/updatedisco.go
```diff
@@ -41,23 +42,24 @@ func UpdateDiscos() (names []string, err error) {
...
 	err = errs.Error()
 	if err != nil {
 		return
 	}
 
 	names = make([]string, 0, len(updated))
 	for filename, _ := range updated {
-		filepath := path.Join(absolutePath, filename)
-		names = append(names, filepath)
+		path := path.Join(absolutePath, filename)
```
tcoffee-google: Name `path` is also an imported module --- maybe `fileAbsolutePath` or `location`?

src/main/common/updatedisco.go
```diff
@@ -120,9 +132,9 @@ func writeDiscoCache(indexData []byte, absolutePath string, directory os.FileInf
 	}
 	size := len(index.Items)
 
-	filepath := path.Join(absolutePath, "index.json")
-	if err := ioutil.WriteFile(filepath, indexData, perm); err != nil {
-		err = fmt.Errorf("Error writing Discovery index to %v: %v", filepath, err)
+	path := path.Join(absolutePath, "index.json")
```
tcoffee-google: As above; here the `path` identifier overlap is particularly confusing.

src/main/common/updatedisco.go
```diff
@@ -164,15 +176,16 @@ func writeDiscoCache(indexData []byte, absolutePath string, directory os.FileInf
...
+// `absolutePath` whose names do not appear in the map of `updated` files, and accumulates any
+// further `errors`.
+func cleanDiscoCache(absolutePath string, filepaths []string, updated map[string]bool, errors *errorlist.Errors) {
+	for _, path := range filepaths {
+		_, filename := filepath.Split(path)
 		if !updated[filename] {
-			filepath := path.Join(absolutePath, filename)
-			if err := os.Remove(filepath); err != nil {
-				errors.Add(fmt.Errorf("Error deleting expired Discovery doc %v: %v", filepath, err))
+			path = filepath.Join(absolutePath, filename)
```
tcoffee-google: Is there a reason to use `filepath.Join` instead of `path.Join`? Right now both are used in this file.

src/main/common/updatedisco.go
```diff
@@ -190,15 +203,15 @@ func UpdateAPI(API apiInfo, absolutePath string, permissions os.FileMode, update
 	fmt.Printf("Updating API: %v %v ...\n", API.Name, API.Version)
 	filename := API.Name + "." + API.Version + ".json"
 	updateChannel <- filename
-	filepath := path.Join(absolutePath, filename)
+	path := path.Join(absolutePath, filename)
```
tcoffee-google: As above.

src/main/common/updatedisco.go
saicheems: That's embarrassing... I renamed them to avoid collision with `filepath` :/

Renamed them all.

src/main/common/updatedisco.go
saicheems: Done

src/main/common/updatedisco.go
saicheems: Done, good call

src/main/common/updatedisco.go
saicheems: Done

src/main/common/updatedisco.go
saicheems: Switched back to `path.Join`

tcoffee-google:
While unlikely, it occurred to me that this surprising property of [`filepath.Glob`](https://golang.org/pkg/path/filepath/#Glob) could produce unexpected results: "Glob ignores file system errors such as I/O errors reading directories. The only possible returned error is ErrBadPattern, when pattern is malformed."

PR #51, opened 2017-06-22
Add Discovery update server

saicheems:


server/main.py
```diff
@@ -0,0 +1,105 @@
...
+
+    with TemporaryDirectory() as tmp_dir:
+        go_dir = os.path.join(tmp_dir, 'go')      # /tmp/go
+        os.makedirs(os.path.join(go_dir, 'src'))  # mkdir -p /tmp/go/src
+
+        # /tmp/discovery-artifact-manager
+        dartman_dir = os.path.join(tmp_dir, 'discovery-artifact-manager')
+
+        _call(('git clone'
+               ' https://github.com/saicheems/discovery-artifact-manager'
```
tcoffee-google: Reminder to switch from fork.

server/main.py
```diff
@@ -0,0 +1,105 @@
...
+    Args:
+        cmd (string): A command to run.
+        check (bool, optional): If true, check_call is run instead of call.
+
+    Returns:
+        int: The return code of the call.
+    """
+    fn = subprocess.call
+    if check:
+        fn = subprocess.check_call
```
tcoffee-google: If this raises `CalledProcessError`, how would we be notified?

server/main.py
```diff
@@ -0,0 +1,105 @@
...
+        #       /tmp/go/src/discovery-artifact-manager
+        _call('ln -s {} {}'.format(
+            os.path.join(dartman_dir, 'src'),
+            os.path.join(go_dir, 'src', 'discovery-artifact-manager')),
+              check=True)
+
+        env = os.environ.copy()
+        env['GOPATH'] = go_dir
+
+        _call('go run src/main/updatedisco/main.go', check=True,
```
tcoffee-google: Using the compiled binary would let you skip all the go installation and plumbing, wouldn't it?

server/main.py
```diff
@@ -0,0 +1,105 @@
...
+              cwd=dartman_dir, env=env)
+
+        _call('git add discoveries', check=True, cwd=dartman_dir)
```
tcoffee-google: Don't we need `git add -A` in case discoveries are removed?

server/main.py
```diff
@@ -0,0 +1,105 @@
...
+                          '/saicheems/discovery-artifact-manager')
+            remote_url = remote_url.format(cred.personal_access_token)
+
+            # Send output to /dev/null so remote_url isn't logged.
+            _call('git remote add github {}'.format(remote_url), check=True,
+                  cwd=dartman_dir, stdout=_DEVNULL, stderr=_DEVNULL)
+            _call('git push github', check=True, cwd=dartman_dir,
+                  stdout=_DEVNULL, stderr=_DEVNULL)
+
+    return ''
```
tcoffee-google: Why empty string? Is this particular to `cron`?

server/main.py
```diff
@@ -0,0 +1,105 @@
...
+        cred = _get_github_credential()
+
+        returncode = _call(
+            ('git -c user.name="{}" -c user.email="{}"'
+             ' -m "Autogenerated Discovery document update"').format(
+                 cred.name, cred.email),
+            cwd=dartman_dir)
+
+        # returncode is non-zero if there's nothing to commit.
+        if not returncode:
```
tcoffee-google: Will the previous `_call` log the error if there's nothing to commit? We want the reason to be obvious if we're not seeing updates.

server/main.py
saicheems: The Flask framework catches the exception, logs the stack trace, and returns a 500 response.

server/main.py
saicheems: Yes, that's a good point, I'll do that now.

saicheems: Oh, I just remembered why I didn't do that --

I'm cloning `discovery-artifact-manager` on each run to `tmp` to ensure there's no possibility of weird git sync conflicts. Since `updatedisco` updates the discovery files under `$GOPATH` (unless I'm mistaken?) I have to point to my cloned repo so those discovery files are updated.

server/main.py
saicheems: I don't think so, as far as I know the add call also stages deleted files. So far, it's worked in practice, it's deleted some files on my fork.

server/main.py
saicheems: Yes, the log will say something like,

"nothing to commit..."

server/main.py
saicheems: It's arbitrary, just an easy way to return an empty 200 response.

server/main.py
```diff
@@ -0,0 +1,105 @@
...
+    if check:
+        fn = subprocess.check_call
+    return fn(shlex.split(cmd), **kwargs)
+
+
+@app.route('/cron/discoveries')
+def cron_discoveries():
+    # This header can't be spoofed, see
+    # https://cloud.google.com/appengine/docs/flexible/nodejs/scheduling-jobs-with-cron-yaml#securing_urls_for_cron
+    if request.headers.get('X-AppEngine-Cron') is None:
```
vchudnov-g: nit: s/`AppEngine`/`Appengine`/ to be consistent with the way it's presented in that link

server/.dockerignore
```diff
@@ -0,0 +1,19 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
```
saicheems: Done

server/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Install Go 1.8
+RUN wget https://storage.googleapis.com/golang/go1.8.linux-amd64.tar.gz
+RUN tar -xvf go1.8.linux-amd64.tar.gz
+RUN mv go /usr/local
+ENV PATH="/usr/local/go/bin:${PATH}"
+
+# Set virtualenv environment variables. This is equivalent to running
+# source /env/bin/activate
+ENV VIRTUAL_ENV /env
```
saicheems: Did the former

server/main.py

saicheems:
I plan to merge and set this up in the afternoon - at that point `googleapis-publisher` will automatically commit to this repository with discovery updates.

server/main.py
tcoffee-google: I believe `updatedisco` updates discovery files to the top-level 'discoveries' directory under whatever repo it's called within.

PR #53, opened 2017-06-28
Update the Google API client generator

saicheems:
Reflects changes made to the source of truth.

### DIFF
--- a/google-api-client-generator/src/googleapis/codegen/cpp_generator.py
+++ b/google-api-client-generator/src/googleapis/codegen/cpp_generator.py
@@ -36,7 +36,6 @@ flags.DEFINE_boolean(
     True,
     'Adds owner subdirectory to --output_dir path.')
 
-
 # This pattern is used to extract the DT and DD text groups.
 # It is meant to be applied to the split partions of the dl_pattern_re.
 # It turns  'A - Description A\nB - Description B'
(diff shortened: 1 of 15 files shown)


bshaffer:
Shouldn't this be pulling in https://github.com/google/apis-client-generator instead of including the code in this repository? Why are we maintaining both?

saicheems:
It had considered it, but `apis-client-generator` has diverged from the internal generator and is not being properly maintained at all, and I wasn't sure If I would be messing up other people's workflows by forcing the latest version.

That's still an option of course, but it seemed more trouble than it was worth at the moment to deal with the merge.

google-api-client-generator/src/googleapis/codegen/cpp_generator.py
```diff
@@ -215,15 +214,18 @@ def AnnotateApiForLanguage(self, the_api):
   def AnnotateApi(self, the_api):
     """Annotate a Api with C++ specific elements."""
```
vchudnov-g: nit: "an API"

saicheems: I won't change anything here, since I'd prefer not to diverge in any way from the source of truth.

saicheems:
This PR contains the latest changes to `google-api-client-generator` since it was last updated from the internal source of truth, none of this code was edited by me.

PR #54, opened 2017-06-28
Add PHP cron path

saicheems:
This commit adds the `/cron/clients/php` route, which is triggered every
30 min to update the `google-api-php-client-services` library.


server/main.py
```diff
@@ -102,5 +105,174 @@ def cron_discoveries():
...
+        # /tmp/discovery-artifact-manager
+        dartman_dir = os.path.join(tmp_dir, 'discovery-artifact-manager')
+        _call(('git clone'
+               ' https://github.com/googleapis/discovery-artifact-manager'
+               ' {}').format(dartman_dir), check=True)
+
+        # /tmp/google-api-php-client-services
+        client_lib_dir = os.path.join(tmp_dir, 'google-api-php-client-services')
+        _call(('git clone'
+               ' https://github.com/saicheems/google-api-php-client-services'
```
bshaffer: this should be formatted to be `account.username`

server/main.py
```diff
@@ -102,5 +105,174 @@ def cron_discoveries():
...
+            new_minor_revision = str(int(minor_revision) + 1)
+            # ex: 'v1.13'
+            new_version = version_re.sub(
+                r'\1.{}'.format(new_minor_revision), latest_tag)
+
+            _call('git tag {}'.format(new_version), check=True,
+                  cwd=client_lib_dir)
+
+            remote_url = ('https://{}:{}@github.com'
+                          '/saicheems/google-api-php-client-services')
```
bshaffer: this should be formatted to be `account.username`

saicheems: Will do, `saicheems` is a placeholder for my fork, where the script is running ATM.

I'll switch it before this is merged to point to `/google/google-api-php-client-services`.

server/main.py
```diff
@@ -102,5 +105,174 @@ def cron_discoveries():
...
+                    os.path.join(service_dir, service_name),
+                    os.path.join(service_dir, service_name)),
+                      check=True)
+                # Copy the newly generated service back.
+                # cp /tmp2/Foo.php /tmp/google-api-php-client-services/src/Google/Service/Foo.php
+                _call('cp {}.php {}'.format(
+                    os.path.join(tmp_dir2, service_name), service_dir),
+                      check=True)
+                # cp -r /tmp2/Foo /tmp/google-api-php-client-services/src/Google/Service/Foo
+                _call('cp -r {} {}'.format(
```
vchudnov-g: Is it possible to combine these two `_call`s into one?

saicheems: I wasn't sure how to do it, since `cp` only takes one source file. Theoretically you could use a wildcard, but I was afraid of unexpected behavior.

saicheems: The other way to do this is to use `shutil`. I'll just do that, it'll be more obvious.

saicheems: Actually I switched it back to `cp`, it's a bit more predictable and less fragile than using `shutil`.

saicheems:
Okay, switched the PR to point to the real repo instead of my fork.

Once this is reviewed and merged, I'll push this version and `googleapis-publisher` will attempt to commit directly to google/google-api-php-client-services

saicheems:
Also changed the cron schedule to update the client library once per day at most.

server/main.py
vchudnov-g: OK. I do suggest that in each of these two `_call`s, you put each of the substituted-in params on a separate line for readability.

saicheems: Will do in follow-up

PR #55, opened 2017-07-13
Split PHP route into update and release steps

saicheems:
Combine the update commits into a single message.


saicheems:
With this commit, the server combines all update commits into one commit, ex:
https://github.com/saicheems/google-api-php-client-services/commit/8c2f5ffd2902a2c3a565cd02c02190b5206ef554

It also splits the tag releasing into a second route, which Brent asked for.

Now, every day at 0:15 updates are pushed to master. Every Monday at 0:45, if there have been any commits since the last tag, a new tag is pushed with the minor version incremented.

server/main.py
```diff
@@ -250,7 +321,7 @@ def cron_clients_php():
             minor_revision = match.group(2)
             # ex: '13'
             new_minor_revision = str(int(minor_revision) + 1)
-            # ex: 'v1.13'
+            # ex: 'v0.13'
             new_version = version_re.sub(
                 r'\1.{}'.format(new_minor_revision), latest_tag)
```
bshaffer: Does this mean we are tagging a new major version?

saicheems: No, I actually thought this might be confusing before, so I'll add a comment explaining what's happening:

`\1' represents the first group in the regex, so we always increment only the minor version.
So if the last tag was `0.12`, we'd release `0.13`, and if the last tag was `1.12`, we'd release `1.12`.

saicheems: This may not be what we always want to do, but I wanted to defend against the possibility of somehow regressing the major version.

PR #58, opened 2017-08-03
Add Node.js cron paths

saicheems:


saicheems:
This commit introduces jobs to update and release the Node.js client library. Since there are more steps involved for this client, the process is a bit more complicated and conservative than for PHP.

##### `update`

1. Clone the client library.
2. Generate, build, and run tests.
3. Commit and push.

##### `release`

1. Clone the client library.
2. Bail if there were no commits since the latest git tag.
3. Bail if any commits since the latest git tag had an author other than `googleapis-publisher@google.com`.
4. Bail if the latest git tag does not match the latest `googleapis` package version on `npm`.
5. Run tests.
6. Increment the major version if any clients were deleted, or the minor version otherwise.
7. Update `package.json` and `CHANGELOG.md`.
9. Commit the changes to `package.json` and `CHANGELOG.md` and release a new tag.
10. Push to master.
11. Publish to `npm`.

saicheems:
FYI, @jmdobry

server/main.py
```diff
@@ -105,6 +118,253 @@ def cron_discoveries():
...
+              cwd=client_lib_dir)
+
+        # Build all clients.
+        _call('node --max_old_space_size=2000 /usr/bin/npm run build',
+              check=True, cwd=client_lib_dir)
+
+        # Run tests.
+        _call('npm run test', check=True, cwd=client_lib_dir)
+
+        # `version_re` matches versions like "20.1.0".
```
vchudnov-g: For readability, I would suggest factoring out into separate functions some of these stanzas that provide inputs to the main role of this method (and same for the other methods). For example, here, `major_version, minor_version, patch_version = get_version(latest_tag)`. Makes the method more readable.

And I think this suggestion becomes very important when there's repeated code.

server/main.py
```diff
@@ -105,6 +118,253 @@ def cron_discoveries():
...
+        statuses = [match.group(1) for match
+                    in name_version_re.finditer(diff_ns.decode('utf-8'))]
+        # `changes` is a map of API ID to file status ("A", "D", or "M").
+        changes = {}
+        for match in name_version_re.finditer(diff_ns.decode('utf-8')):
+            status = match.group(1)
+            name_version = '{}:{}'.format(match.group(2), match.group(3))
+            changes[name_version] = status
+
+        # If any clients were deleted, increment the major version.
```
vchudnov-g: In either case, `patch_version = 0`, right?

server/main.py
```diff
@@ -105,6 +118,253 @@ def cron_discoveries():
...
+        diff_ns = subprocess.check_output(
+            shlex.split(
+                'git diff --name-status {}..HEAD --oneline'.format(
+                    latest_tag)),
+            cwd=client_lib_dir)
+
+        # Get the status for each file.
+        statuses = [match.group(1) for match
+                    in name_version_re.finditer(diff_ns.decode('utf-8'))]
+        # `changes` is a map of API ID to file status ("A", "D", or "M").
```
vchudnov-g: Since you're processing the file status here as you did in the previous function, consider whether creating a helper that returns `added, deleted, updated, changes = modifiedFiles(...)` for use in both places might simplify things or make the code more readable. (It may not, which is fine.)

server/main.py
```diff
@@ -105,6 +118,253 @@ def cron_discoveries():
...
+        _call(cmd, check=True, cwd=client_lib_dir)
+        _call('git tag {}'.format(new_version), check=True,
+              cwd=client_lib_dir)
+
+        remote_url = ('https://{}:{}@github.com'
+                      '/google/google-api-nodejs-client')
+        remote_url = remote_url.format(github_account.username,
+                                       github_account.personal_access_token)
+
+        # Send output to /dev/null so `remote_url` isn't logged.
```
vchudnov-g: My style suggestion(feel free to disagree): 
```
runGitCommand('git remote add github {}'.format(remote_url))
runGitCommand('git push github')
runGitCommand('git push github --tags')
```

(and you can also use this helper function in the other method).

server/main.py
```diff
@@ -105,6 +118,253 @@ def cron_discoveries():
...
+
+        # Commit the changes to `package.json` and `CHANGELOG.md`.
+        cmd = 'git -c user.name="{}" -c user.email="{}" commit -a -m "{}"'
+        cmd = cmd.format(github_account.name, github_account.email,
+                         new_version)
+        _call(cmd, check=True, cwd=client_lib_dir)
+        _call('git tag {}'.format(new_version), check=True,
+              cwd=client_lib_dir)
+
+        remote_url = ('https://{}:{}@github.com'
```
vchudnov-g: These are repeated from the previous method, right? Factor out into a helper function.

server/main.py
```diff
@@ -307,41 +568,42 @@ def cron_clients_php_release():
...
+                      '/google/google-api-php-client-services')
+        remote_url = remote_url.format(account.username,
+                                       account.personal_access_token)
+
+        # Send output to /dev/null so `remote_url` isn't logged.
+        _call('git remote add github {}'.format(remote_url), check=True,
```
vchudnov-g: Same here with the helper method I suggest above.

server/main.py
```diff
@@ -307,41 +568,42 @@ def cron_clients_php_release():
...
+        # '\1' is substituted with the first group captured by the
+        # `version_re` pattern. This replacement defends against the
+        # possibility of regressing the major version.
+        # ex: 'v0.13'
+        new_version = version_re.sub(
+            r'\1.{}'.format(new_minor_version), latest_tag)
+        _call('git tag {}'.format(new_version), check=True,
+              cwd=client_lib_dir)
+
+        remote_url = ('https://{}:{}@github.com'
```
vchudnov-g: If these are used more than once, factor out.

server/Dockerfile
```diff
@@ -1,21 +1,27 @@
...
 # Set virtualenv environment variables. This is equivalent to running
 # source /env/bin/activate
 ENV VIRTUAL_ENV /env
 ENV PATH /env/bin:$PATH
 ADD requirements.txt /app/
 RUN pip install -r requirements.txt
 ADD . /app/
-# 10 minute timeout.
-CMD gunicorn -b :$PORT main:app --timeout 600 --workers 4
+# 20 minute timeout.
```
saicheems: This is to prevent `gunicorn` from killing tasks before they finish. Some of the jobs take longer than 10 min start to finish.

server/Dockerfile
```diff
@@ -1,21 +1,27 @@
...
 RUN tar -xvf go1.8.linux-amd64.tar.gz
 RUN mv go /usr/local
 ENV PATH "/usr/local/go/bin:${PATH}"
 
+# Install Node.js 8.x.
+RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
+RUN apt-get install -y nodejs
+
+# Install Python 3.6.
+RUN virtualenv /env -p python3.6
```
saicheems: I don't think it's needed here, doesn't really save any lines or simplify.

server/Dockerfile
saicheems: I'll add a comment to note.

jmdobry:
LGTM

server/main.py
saicheems: Yeah, I removed it.

server/main.py
saicheems: I've created another option, `quiet`, instead, that redirects all output to `/dev/null` if `True`.

server/main.py

server/main.py

server/main.py

server/main.py
saicheems: I'm not sure it helps much with readability in this case :/

Since the regexes between repos changes, and the formatting of the api name/version combo changes... I don't think there's much saved by generalizing these few lines.

server/main.py
```diff
@@ -105,6 +118,253 @@ def cron_discoveries():
...
+            raise Exception(
+                'latest tag does not match the pattern \'{}\': {}'.format(
+                    version_re.pattern, latest_tag))
+
+        major_version = int(match.group(1))
+        minor_version = int(match.group(2))
+        patch_version = int(match.group(3))
+
+        # `name_version_re` matches strings like
+        # "M    apis/foo/v1.ts"
```

server/main.py
saicheems: I agree... but let me come back to that in a follow up to more generally refactor the layout of this file.

The trick is that every repo has it's own versioning style that isn't necessarily semantic versioning, so every `get_version` function would have to be different. (ex, PHP: `v0.19`, Node.js: `15.1.0`).

I think the better way to do this is to have a file for each language, or a class for each language, and have private methods to deal with how a specific client lib deals with versioning and stuff.

PR #61, opened 2017-09-12
Add accounts module

saicheems:
Contains definitions and getters for the various accounts the server
needs access to.

Account information is retrieved from Datastore.

### DIFF
--- a/server/tasks/accounts.py
+++ b/server/tasks/accounts.py
@@ -22,9 +22,10 @@ NpmAccount = namedtuple('NpmAccount', 'auth_token')
 RubyGemsAccount = namedtuple('RubyGemsAccount', 'api_key')
 
 
-def _obj(kind):
+def _get(type_):
     client = datastore.Client()
-    return list(client.query(kind=kind).fetch())[0]
+    obj = list(client.query(kind=type_.__name__).fetch())[0]
+    return type_(*[obj[x] for x in type_._fields])
 
 
 def get_github_account():
@@ -33,9 +34,7 @@ def get_github_account():
     Returns:
         GitHubAccount: a GitHub account.
     """
-    obj = _obj('GitHubAccount')
-    return GitHubAccount(obj['name'], obj['email'], obj['username'],
-                         obj['personal_access_token'])
+    return _get(GitHubAccount)
 
 
 def get_npm_account():
@@ -44,8 +43,7 @@ def get_npm_account():
     Returns:
         NpmAccount: an npm account.
     """
-    obj = _obj('NpmAccount')
-    return NpmAccount(obj['auth_token'])
+    return _get(NpmAccount)
 
 
 def get_rubygems_account():
@@ -54,5 +52,4 @@ def get_rubygems_account():
     Returns:
         RubyGemsAccount: a RubyGems account.
     """
-    obj = _obj('RubyGemsAccount')
-    return RubyGemsAccount(obj['api_key'])
+    return _get(RubyGemsAccount)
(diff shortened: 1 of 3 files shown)


saicheems:
For some background, the full implementation of the server will be split in to many modules. The common functionality is implemented in modules like `_check_output`, `_git`, `accounts`...

Functions for updating and releasing client libraries are implemented in the modules `google_api_go_client`, `google_api_php_client_services`... (not yet put up on GitHub, at the time of this comment.)

server/tasks/accounts.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def get_github_account():
+    """Returns the GitHub account stored in Datastore.
+
+    Returns:
+        GitHubAccount: a GitHub account.
+    """
+    obj = _obj('GitHubAccount')
+    return GitHubAccount(obj['name'], obj['email'], obj['username'],
```
vchudnov-g: Why do you need the `namedtuple`s if `obj` already has a mapping of field names to their values? Are you just trying to abstract away any storage implementation details?

server/tasks/accounts.py
```diff
@@ -0,0 +1,58 @@
...
+                         obj['personal_access_token'])
+
+
+def get_npm_account():
```
vchudnov-g: All these functions repeat the field names and are themselves very similar.

How about making a common base class `AccountData` which has a `namedtuple` field (call it `details`) and a function `Get()` that does the `_obj()` call and iterates over `details._fields` and provides an accessor for each? Then `GitHubAccount`, `NpmAccount` and `RubyGemsAccount` can each be specializations of `AccountData` for each `kind` and set of fields.

server/tasks/accounts.py
saicheems: Yes, I think it's more apparent what an object is if a schema is enforced via a `namedtuple` or `class`.
It's also easier to document in a function signature than handing a dict around and explaining what keys it holds.

server/tasks/accounts.py
saicheems: I'm a little confused by what you mean,

```
def _obj(kind):
    pass

class AccountData(object):
    def __init__(self):
        pass

    def github(self):
        return AccountData(fields={...})
```

^ like that?
My hesitation is that introducing a hierarchy in this module might be needlessly verbose. For a n < 4 field object, using a `namedtuple` (immutable) seems sufficient.

vchudnov-g: My concern is that you're repeating the field names when you could simply have them in one place (and to a lesser extent, that all three `get_*` functions are doing essentially the same thing). Here's another scheme to address that which is simpler as it doesn't require inheritance. I haven't tested that all this magic works in Python, but I _think_ it should.

```
# account is one of the *Account namedtuples
def get_account(account):
  kind = account.__name__
  obj = _obj(kind)

  fields = []
  for f in account._fields: 
    fields.append(obj[f])  
  return account(*fields)  # not sure if this instantiation will work as is

# sample usage
myaccount = get_account(GitHubAccount)
```

This would replace the three `get_*` functions with just this one.

server/tasks/accounts.py
vchudnov-g: sg

server/tasks/accounts.py
saicheems: PTAL

I simplified the logic to use essentially what you suggested, so there's no guessing of or repetition of field names.

It is my personal preference not to have functions return different types... so I'm still keeping the three `get_*_account()` function separate. I think it's more obvious what the interface is that way, but let me know if you strongly prefer that all the functions be condensed.

vchudnov-g: Works for me.

You're still returning varying types from the module-private function `_obj`, but not exposing a function that returns varying types. That's fine.

(and hooray for list comprehensions!)

PR #62, opened 2017-09-12
Add _check_output module

saicheems:
Contains a convenient wrapper over `subprocess.check_output` which
raises a nice exception containing stdout and stderr information on
failure.

### DIFF
--- a/server/tasks/_check_output.py
+++ b/server/tasks/_check_output.py
@@ -1,3 +1,16 @@
+# Copyright 2017, Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 """Contains a convenient wrapper over `subprocess.check_output`."""
 from subprocess import PIPE, Popen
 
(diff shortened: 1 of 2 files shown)


server/tasks/_check_output.py
```diff
@@ -0,0 +1,37 @@
+"""Contains a convenient wrapper over `subprocess.check_output`."""
```
vchudnov-g: Licenses in all files.

server/tasks/_check_output.py
```diff
@@ -0,0 +1,37 @@
...
+
+def _call(args, **kwargs):
+    proc = Popen(args, stdout=PIPE, stderr=PIPE, **kwargs)
+    stdoutdata, stderrdata = proc.communicate()
+    return (proc.returncode, stdoutdata.decode('utf-8'),
+            stderrdata.decode('utf-8'))
+
+
+def check_output(args, **kwargs):
+    """A convenient version of `subprocess.check_output`.
```
vchudnov-g: Suggested enhancement for readability:
Split `args[0]` on spaces and replace it with the resulting list. This allows  common invocations like `check_output(["dosomething myarg yourarg"])`

For situations where the arguments may themselves contain spaces, this still allows the current usage: `check_output(["dosomething myarg"] ["your arg"])`

saicheems: The reason I switched from `shlex.split` to the arg style signature was for security. Since we're pulling names from the filesystem and from documents downloaded online, I was concerned about injection of escape characters into the subprocess calls.

Ex: By some freak occurrence, the id of an API gets changed to `"; bash -- curl ...`.

I think it's better to err on the side of safety even if there's some readability loss (I argue it's not too big a deal). Even if the code at the moment do string replaces on `arg[0]`, someone in the future might without realizing.

server/tasks/_check_output.py
saicheems: Done

PR #63, opened 2017-09-12
Add _commit_message module

saicheems:
Contains a function which builds a nice, standard commit message for an
autogenerated update.


server/tasks/_commit_message.py
```diff
@@ -0,0 +1,35 @@
+"""Contains a function which builds a nice commit message."""
```
vchudnov-g: License comment in all files.

server/tasks/_commit_message.py
```diff
@@ -0,0 +1,35 @@
...
+from datetime import date
+
+
+def build(added, deleted, updated, subject=None):
+    """Returns a nice commit message.
```
vchudnov-g: Here and above, could we use something more descriptive than "nice"? "Comprehensive", "detailed", "complete", "consistent",....

server/tests/test__commit_message.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_build_subject():
+    commitmsg = _commit_message.build(None, None, None, subject='X')
+    assert commitmsg == 'X'
+
+
+def test_build_added_subject():
+    added = {'foo:v1'}
+    commitmsg = _commit_message.build(added, None, None, subject='X')
+    assert commitmsg == 'X\n\nAdd:\n- foo:v1'
```
vchudnov-g: For readability here and below, consider using pseudo-multiline strings a la
```
assert commitmsg == ("X\n"
                     "\n"
                     "Add:\n"
                     "- foo:v1")
```

server/tasks/_commit_message.py
saicheems: Done

server/tasks/_commit_message.py
saicheems: Changed to "detailed"

server/tests/test__commit_message.py
saicheems: Done

PR #64, opened 2017-09-13
Add _git module

saicheems:
Contains convenience functions and classes for cloning and operating on
Git repositories.

### DIFF
--- a/server/tasks/_git.py
+++ b/server/tasks/_git.py
@@ -1,0 +1,224 @@
+# Copyright 2017, Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Implements a utility for cloning and operating on Git repositories."""
+import re
+from enum import Enum
+
+from tasks._check_output import check_output
+
+_DIFF_NAME_STATUS_RE = re.compile(r'^(\w)\s+(.*)$')
+
+Status = Enum('Status', 'ADDED DELETED UPDATED UNKNOWN')
+
+
+def clone_from_github(path, dest, github_account=None):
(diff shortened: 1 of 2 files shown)


server/tasks/_git.py
```diff
@@ -0,0 +1,224 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Implements a utility for cloning and operating on Git repositories."""
+import re
+from enum import Enum
+
+from tasks._check_output import check_output
+
+_DIFF_NAME_STATUS_RE = re.compile(r'^(\w)\s+(.*)$')
```
vchudnov-g: comment

server/tasks/_git.py
```diff
@@ -0,0 +1,224 @@
...
+            list(str): a list of emails of the authors of all commits since
+                `rev`.
+        """
+        data = check_output(
+            ['git', 'log', '{}..HEAD'.format(rev), '--pretty=format:"%ae"'],
+            cwd=self.filepath)
+        return data.strip().split('\n')
+
+    def checkout(self, branch):
+        """Switch branches.
```
vchudnov-g: "Switches"

server/tasks/_git.py
```diff
@@ -0,0 +1,224 @@
...
+        Args:
+            branch (str): the name of the branch to checkout.
+
+        Raises:
+            CallError: if the call returns a non-zero return code.
+        """
+        check_output(['git', 'checkout', branch], cwd=self.filepath)
+
+    def commit(self, message, name, email):
+        """Record changes to the repository.
```
vchudnov-g: "Records"

server/tasks/_git.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        check_output(
+            ['git',
+             '-c', 'user.name={}'.format(name),
+             '-c', 'user.email={}'.format(email),
+             'commit', '-a', '--allow-empty-message', '-m', message],
+            cwd=self.filepath)
+
+    def diff_name_status(self, rev=None, staged=True):
+        """Return a list of status, filename pairs of changes from `rev`.
```
vchudnov-g: "Returns"

saicheems:
So I spoke with Ethan some time ago about making parts of artman and dartman compatible... and I suggested that we refactor in the future, once this server is at 100%.

I think if it becomes a concern that we repeat a lot of common code (I don't think it should be too big an issue, outside of this git stuff), we should refactor a "googleapis/artman-common" python module that hosts those modules.

Probably not worth the effort right away though.

PR #66, opened 2017-09-15
Add google_api_go_client module

saicheems:
Contains functions for updating the google-api-go-client repository.


server/tasks/google_api_go_client.py
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Contains update/release functions for google-api-go-client."""
+import re
+from datetime import date
+
```
vchudnov-g: Why are these in two sections?

server/tasks/google_api_go_client.py
```diff
@@ -0,0 +1,66 @@
...
+def update(filepath, github_account):
+    """Updates the google-api-go-client repository.
+
+    Args:
+        filepath (str): the directory to work in.
+        github_account (GitHubAccount): the GitHub account to commit with.
+    """
+    env = os.environ.copy()
+    env['GOPATH'] = filepath
+    check_output(['go', 'get', '-d', '-t', 'google.golang.org/api/...'],
```
vchudnov-g: nit: I like my previous suggestion of having check_output split the first element of the array on spaces to increase readability in cases like this.

server/tasks/google_api_go_client.py
saicheems: It's a Python style thing, I guess there are a few schools of thought on import order, but pylint at least complains if `import os` and `from os import ...` aren't grouped together.

server/tasks/google_api_go_client.py
saicheems: While I agree that readability might improve, I do think that doing what you suggest could also make this function more confusing for future maintainers... having only the first element of the list be split is inconsistent, and non-standard compared to `subprocess.check_output`.

I think it's better in this case to accept it as-is. Having each arg separate also ensures the tests capture exactly what gets called (can know that `['go', 'get', ...]` is actually being sent to `Popen`, and not `['go get ..., 'foo', 'bar']`, somehow).

PR #68, opened 2017-09-18
Add google_api_ruby_client module

saicheems:
Contains functions for updating and releasing the
google-api-ruby-client repository.


server/tasks/google_api_ruby_client.py
```diff
@@ -0,0 +1,195 @@
...
+    added, deleted, updated = set(), set(), set()
+    for filename, status in repo.diff_name_status(staged=False):
+        match = _SERVICE_FILENAME_RE.match(filename)
+        if not match:
+            continue
+        {
+            _git.Status.ADDED: added,
+            _git.Status.DELETED: deleted,
+            _git.Status.UPDATED: updated
+        }.get(status, set()).add(match.group(1))
```
vchudnov-g: I think it would be clearer if you define the map outside the loop, after you create the empty sets, rather than creating a map in every iteration using this cool this way.

server/tasks/google_api_ruby_client.py
```diff
@@ -0,0 +1,195 @@
...
+    check_output(['bundle', 'exec', 'rake', 'spec'], cwd=repo.filepath)
+
+
+def _update_changelog_md(repo, new_version, added, deleted, updated):
+    data = '# {}\n'.format(new_version)
+    if deleted:
+        data += '* Breaking changes:\n'
+    for name_version in sorted(deleted):
+        data += '  * Deleted `{}`\n'.format(name_version)
+    if added or updated:
```
vchudnov-g: Could some updates be backwards incompatible? Or is there a convention to not author those?

server/tasks/google_api_ruby_client.py
```diff
@@ -0,0 +1,195 @@
...
+    diff_ns = repo.diff_name_status(rev=latest_tag)
+    for filename, status in diff_ns:
+        match = _SERVICE_FILENAME_RE.match(filename)
+        if not match:
+            continue
+        {
+            _git.Status.ADDED: added,
+            _git.Status.DELETED: deleted,
+            _git.Status.UPDATED: updated
+        }.get(status, set()).add(match.group(1))
```
vchudnov-g: As above: define map outside the loop

server/tasks/google_api_ruby_client.py
```diff
@@ -0,0 +1,195 @@
...
+        return
+    _run_tests(repo)
+    commitmsg = _commit_message.build(added, deleted, updated)
+    repo.add(['api_names_out.yaml', 'generated'])
+    repo.commit(commitmsg, github_account.name, github_account.email)
+    repo.push()
+
+
+def release(filepath, github_account, rubygems_account):
+    """Releases a new version in the google-api-ruby-client repository.
```
vchudnov-g: Add a few words about how releasing works and how it differs from updates. Think of somebody else suddenly tasked with maintaining these libraries.

server/tasks/google_api_ruby_client.py
saicheems: Technically it is possible, but it's the kind of thing that is impossible to deal with in practice, short of only releasing major versions.

In this case, we have to trust that API owners aren't releasing breaking changes on new revisions of the same version. That is a rule they shouldn't be violating anyway.

server/tasks/google_api_ruby_client.py
saicheems: Done

PR #70, opened 2017-09-21
Add force option to release functions

saicheems:
Provides a way of releasing versions using the bot even if another
author has commited since the last tag.


server/tests/test_google_api_nodejs_client.py
```diff
@@ -401,6 +401,66 @@ def test_release_different_authors_since_latest_tag(clone_from_github_mock):
...
+@patch('tasks.google_api_nodejs_client.check_output')
+@patch('tasks.google_api_nodejs_client._git.clone_from_github')
+def test_release_force(clone_from_github_mock,
+                       check_output_mock,
+                       open_mock,
+                       date_mock,
+                       expanduser_mock):
+    repo_mock = Mock()
+    repo_mock.latest_tag.return_value = '20.1.0'
+    repo_mock.authors_since.return_value = ['test@test.com', 'test2@test.com']
```
vchudnov-g: nit: in this file and the other, could you use 'author@test.com', 'author2@test.com' or "alice@test.com", "bob@test.com"?  Make it more obvious (to me at least) you're simulating a different contributor.

server/tasks/google_api_nodejs_client.py
```diff
@@ -200,13 +200,17 @@ def release(filepath, github_account, npm_account):
...
     repo = _git.clone_from_github(
         _REPO_PATH, join(filepath, _REPO_NAME), github_account=github_account)
     latest_tag = repo.latest_tag()
     version = _Version(latest_tag)
     authors = repo.authors_since(latest_tag)
-    if not authors or not all([x == github_account.email for x in authors]):
+    if not authors:
+        return
+    if not force and not all([x == github_account.email for x in authors]):
         return
```
vchudnov-g: So everything in this function up to this line seems common among all the languages. Could you factor out into common utility function that returns the multiple values you're setting, and a "continue" flag?

saicheems: I created a common module to do the author checks, I haven't come up with a real nice way to return all the variables above though (`repo`, `latest_tag`, and `version`). I think it's better to be explicit for those, but if you have an idea lmk.

server/tests/test_google_api_nodejs_client.py
saicheems: Done

server/tasks/_common.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Returns:
+        bool: whether or not to proceed with a release.
+    """
+    authors = repo.authors_since(latest_tag)
+    if not authors:
+        return False
+    if force:
+        return True
+    if not all([x == github_account.email for x in authors]):
```
vchudnov-g: nit: You could combine this if with the fall-through case. In fact you could combine all the ifs:
```
return authors and  \
  (force or \
   all(...))
```

server/tests/test_google_api_nodejs_client.py
vchudnov-g: Alice and Bob are so busy!

server/tasks/google_api_nodejs_client.py
vchudnov-g: Seems like you could at least wrap and return `latest_tag = repo.latest_tag()` in your new function. 

I think `version` should also be doable if, for example, the new function is a method in a parent class of a class that implements version. I think it would be more elegant, but I'm not sure the refactoring is worth the effort right now.

saicheems: I see what you're saying, I'll stick with this for this PR and refactor `_Version` in a follow-up.

PR #71, opened 2017-09-21
Rewrite routes and update cron.yaml/Dockerfile

saicheems:
- Rewrite the routes in `main.py` to call the release/update tasks.
- Rewrite the `cron.yaml` rules to include all release/update tasks.
- Include all required runtimes in `Dockerfile`.
  - Add Node.js, PHP, Python, and Ruby environments.


saicheems:
Best way to review `main.py` might be to view the whole file and transfer comments. The diff is funky because it's been totally rewritten.

vchudnov-g:
Could you provide a more useful PR name.description than "update main.py"?

There are also changes to cron.yaml and Dockerfile---not just import changes. Did you mean to include them?

The .py file is re-written by hand? Or just stuff moved around?

saicheems:
Updated the description.
The py file is totally rewritten, since all the old logic got moved to the `tasks` module.
The changes to `cron.yaml` and `Dockerfile` are intentional, I'm adding all the extra stuff required at runtime to deploy the server/run all tasks.

server/main.py
```diff
@@ -1,348 +1,103 @@
-import glob
-import json
-import os
-import re
-import shlex
-import subprocess
-from collections import namedtuple
-from datetime import date
 from tempfile import TemporaryDirectory
```
vchudnov-g: Copyright
File comments

server/main.py
```diff
@@ -1,348 +1,103 @@
...
-    Returns:
-        int: The return code of the call.
-    """
-    func = subprocess.call
-    if check:
-        func = subprocess.check_call
-    return func(shlex.split(cmd), **kwargs)
-
 
 @app.route('/cron/discoveries')
```
vchudnov-g: Consider for consistency and conciseness:

```
@app.route('/cron/clients/<string:lang>/<string:action>
@verify_cron_header
@get_github_account
def cron_clients_handler(lang, action,github_account):
  with TemporaryDirectory as filepath:
    clients_handler[lang][action].Do(github_account, filepath)
  return ''

clients_handler = {"nodejs": {"update": update_nodejs, "release" : release_nodejs}, \
  # ...
}
```
This new common functionality could be in functions called inside `cron_clients_handler` instead of decorators outside.
The new functions/decorators could also be used elsewhere, as in '/cron/discoveries', for example.

Consider also having the `google_api_LANG_client.ACTION` functions all have the same signature, so that you don't need the adapter functions like the `update_nodejs` in my example. For example,
* maybe get the rubygems account inside the Ruby release method
* always pass to `Do()` above a function closure that can be invoked to get the discovery documents, even though it only needs to be invoked by the PHP release process (and tests! This can simplify unit tests!) 

Ref: http://exploreflask.com/en/latest/views.html#built-in-converters

saicheems: I applied 90% of your advice --

- Added adapter functions for the few update/release functions which required extra args.
- I didn't turn the `github_account` variable into a decorator... I'm not a huge fan of using a decorator just to inject a single variable into a function. I'm not a Python language or style expert though, so if you feel strongly about it I'll change that.
- Split the update/release pathways, since release functions use an extra arg.

server/main.py
```diff
@@ -1,349 +1,104 @@
...
+    """Wrapper over the Ruby release function to standardize parameters."""
+    rubygems_account = accounts.get_rubygems_account()
+    google_api_ruby_client.release(
+        filepath, github_account, rubygems_account, force=force)
+
+
+@app.route('/cron/clients/<string:lang>/update')
+@verify_cron_header
+def cron_clients_lang_update(lang):
+    update = {
```
vchudnov-g: Consider that if the `.release` and `.update` sets of functions each had the same signatures for all their members (ie each set had a common interface), you could have a module-global map

```
handler = {
  "php": {
    "client": google_api_php_client_services,
    "discovery": get_discovery_doc
  },
  "nodejs": {
    "client": google_api_nodejs_client,
    "account": accounts.get_npm_account,
  },
  "ruby": {
    "client": google_api_ruby_client,
    "account": accounts.get_rubygems_account,
  },
  "go": {
    "client": google_api_go_client,
  }
```
and only need to define a new `get_discover_func` function. You could then use it here as

```
client = handler.get(lang, None)
if not client:
  abort(404)
ddocs = client.get("discovery", None)
github_account = ...
with ... as filepath:
  client.update(filepath, github_account, ddocs)
```
and in release as
```
client = handler.get(lang, None)
if not client:
  abort(404)
account = client.get("account", None)
github_account = ...
with ... as filepath:
  client.release(filepath, github_account, account, force=force)
```
I think this makes the handlers shorter, imposes a common interface on the same functionality across languages, can simplify unit tests, and puts all the information about each language route together in one place within this file.

server/Dockerfile
```diff
@@ -1,21 +1,45 @@
...
+RUN apt-get install -y python-pip
+RUN pip install virtualenv
+RUN virtualenv /env -p python3
+
+# Install Ruby 2.3 and Bundler.
+RUN apt-get install -y ruby ruby-dev
+RUN gem install bundler --no-ri --no-rdoc
 
 # Set virtualenv environment variables. This is equivalent to running
 # source /env/bin/activate
```
vchudnov-g: Curious: why not just run `activate` then?

server/cron.yaml
```diff
@@ -3,10 +3,42 @@ cron:
...
+  schedule: every day 00:10
+
+- description: release google/google-api-nodejs-client
+  url: /cron/clients/nodejs/release
+  schedule: every monday 01:10
+
+- description: release (force) google/google-api-nodejs-client
+  url: /cron/clients/nodejs/release?force=true
+  schedule: every 99999 hours  # 11 years.
+
```
vchudnov-g: Why do we have the 11-year "force" entries?

server/Dockerfile
saicheems: Not immediately sure tbh, I copied this bit from the example Dockerfile for a AppEngine flex custom Python app.

If I had to guess, it's that using `source` doesn't carry the configuration across future commands. Dockerfiles aren't built by running commands in sequence if I'm not mistaken, but building multiple intermediate containers in which each command is run. It might be the case that using `ENV` carries the variable into future commands.

server/cron.yaml
saicheems: It's the maximum period of time I can schedule. Anything above `99999` is a parse error.

I figured either this app or the google-api-*-clients wouldn't be around for 11 more years :)

Really it's just to make sure they're not frequently run, the force entries are useful to ignore a check if someone pushed a README change but wants to start the bot up again.

vchudnov-g: So is the idea that if somebody wants to run a force check, they just need to restart the bot and this will be run on startup but not again for 11 years, or until the bot is restarted again? 

I guess I'm asking why we have these entries at all. Is it because they do run at start time, or for documentation?

server/main.py
saicheems: I realize that there are ways to reduce the # of LoC in this routes file by unifying function signatures and being clever with function signature dicts, but I'm just not concerned about the complexity of this ~80 line routes file.

I'm also not convinced that unifying function signatures simplifies tests in any way. The tests are already complicated because of the number of internet and filesystem calls that have to be mocked, dropping arguments only require another module that has it's own tests to be mocked (`tasks.accounts`).

IMO it's better in this kind of case to be very explicit even if there's a little code repetition. I'd prefer to see 20 instances of `with TemporaryFile() as tempfile:` than a dict which very cleverly organizes function handlers and requires the maintainer to conform to an established function signature pattern.

vchudnov-g: sg. I'm not going to block on this.

server/Dockerfile
vchudnov-g: Sounds like a plausible guess. It would be nice to know for sure.

server/cron.yaml
vchudnov-g: I suggest adding an explanation as a comment here.

server/Dockerfile
saicheems: I'll give it a shot without and add a comment.

server/cron.yaml
saicheems: Got it, I'll add some comments.

saicheems: They don't run at start, presumably the first time they'll run is 11 years from the most recent cron.yaml push.

The reason I added them was because occasionally a PR is merged, or a README change is pushed that requires me to go and make a new release manually. It's annoying to do if there was no major change or bug fix, so it's easy to go to the cron panel and use the force release route.

server/Dockerfile
saicheems: Yeah has to be separate, the environment doesn't carry over.

server/cron.yaml
saicheems: Added a comment

PR #72, opened 2017-09-22
Add the git cookie refresh daemon

saicheems:
This script periodically refreshes the
"$HOME/.git-credential-cache/cookie" file with an OAuth2 token obtained
from the Compute service for this app's service account.

Authentication by this method is required for direct pushes to the Go
googlesource repository, which is configured to accept unreviewed
changes to master from this app's service account.


saicheems:
This script was adapted from this source:

https://gerrit.googlesource.com/gcompute-tools/+/master/git-cookie-authdaemon

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,144 @@
...
+import cookielib
+import json
+import os
+import subprocess
+import sys
+import time
+import urllib2
+
+REFRESH = 25  # Seconds remaining when starting refresh.
+RETRY_INTERVAL = 5  # Seconds between retrying a failed refresh.
```
vchudnov-g: s/between/before/

saicheems:
Since the google-api-go-client repository is hosted on googlesource, there's a different authentication mechanism required to enable direct push access. We've configured the google-api-go-client repository to accept direct pushes from this project's service account.

The service account is an @google account authenticated by the compute service, so we can query the compute service for an access token.

In order for pushes to be accepted by googlesource, the git push has to be authenticated, so this script periodically writes an oauth2 token obtained from the compute service for the project's service account to `~/.git-credential-cache/cookie`.

This script is started when the instance starts as part of the `CMD` in `Dockerfile`: https://github.com/saicheems/discovery-artifact-manager/blob/9b760d90475bd80194956f97129734f4fb2bfee8/server/Dockerfile#L44

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,144 @@
...
+            if os.path.exists(p):
+                os.remove(p)
+
+
+def refresh_loop(expires):
+    atexit.register(cleanup)
+    expires = expires - REFRESH
+    while True:
+        now = time.time()
+        expires = max(expires, now + RETRY_INTERVAL)
```
vchudnov-g: Why do you need `+ RETRY_INTERVAL` here? This would delay the time of the first update attempt, right? `acquire_token()` already waits for `RETRY_INTERVAL` on failure.

vchudnov-g: For clarity, I would rename the `expires` on the LHS and in the loop to `update_time`. The `expires` after the loop stays the same, of course.

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,144 @@
...
+    else:
+        os.mkdir(dir_, 0o700)
+    subprocess.call(
+        ['git', 'config', '--global', 'http.cookiefile', COOKIE_JAR])
+
+
+def acquire_token(retry):
+    while True:
+        try:
+            with read_meta('token') as d:
```
vchudnov-g: is it more idiomatic this way rather than `return json.load(read_meta('token'))`?

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,144 @@
...
+
+    cj.save()
+    os.rename(tmp_jar, COOKIE_JAR)
+    return expires
+
+
+def cleanup():
+    if COOKIE_JAR:
+        for p in [COOKIE_JAR, COOKIE_JAR + '.lock']:
+            if os.path.exists(p):
```
vchudnov-g: I'm surprised there isn't a way to delete without having to check for existence first. The closest I found was [this](https://stackoverflow.com/a/27045091).

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,144 @@
...
+import os
+import subprocess
+import sys
+import time
+import urllib2
+
+REFRESH = 25  # Seconds remaining when starting refresh.
+RETRY_INTERVAL = 5  # Seconds before retrying a failed refresh.
+
+META_URL = ('http://metadata.google.internal/computeMetadata/v1/instance'
```
vchudnov-g: Comment what this is used for

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,144 @@
...
+            '/service-accounts/default/')
+COOKIE_JAR = None
```
vchudnov-g: Comment. For example, it's not obvious without looking at the usage that this will eventually hold a filename.

server/scripts/git-cookie-authdaemon
saicheems: In Python 3 there is, but not in Python 2.

server/scripts/git-cookie-authdaemon
saicheems: Seems OK to me, it's pretty explicit what's going on.

server/scripts/git-cookie-authdaemon
saicheems: I'm not sure, I didn't write this file.

On inspection, it looks like `expires` will generally always be the longer interval. I think the only case that the latter (`now + RETRY_INTERVAL`) is greater is if the very first update attempt (called before the loop on line 125) failed.

saicheems: named it `updates` to be consistent.

server/scripts/git-cookie-authdaemon
saicheems: done

server/scripts/git-cookie-authdaemon
saicheems: done

server/scripts/git-cookie-authdaemon
vchudnov-g: too verbose for my liking but whatevs, it's still clear and localized. :)

server/scripts/git-cookie-authdaemon
```diff
@@ -0,0 +1,151 @@
...
+import urllib.parse
+import urllib.request
+
+REFRESH = 25  # Seconds remaining when starting refresh.
+RETRY_INTERVAL = 5  # Seconds before retrying a failed refresh.
+
+# URL to query for an OAuth2 token for the instance's default service account.
+META_URL = ('http://metadata.google.internal/computeMetadata/v1/instance'
+            '/service-accounts/default/')
+# Holds the cookie file under "~/.git-credential-cache/cookie".
```
vchudnov-g: "file name"

saicheems: Done

PR #73, opened 2017-10-03
Fix issues found in live environment

saicheems:
- Some update/release steps failed because of instance memory                   
   constraints. 
- `git log foo..HEAD --pretty=format:"%ae"` was printing emails enclosed
  in double quotes, which were not stripped before comparing to the
  non-quoted email returned from Datastore.
- The output of `npm view googleapis version` had trailing whitespace.
- The latest docs generated in the Node.js release step were not being copied
  to the `latest` folder.
- `gem search` does not support the `-e` flag in Ruby 2.3.
- The `discovery_v1` client library was missing before the call to
  `./script/generate` in the Ruby update step.
- `version.rb` was not opened with write permissions before being
  written to in the Ruby release step.


server/cron.yaml
```diff
@@ -1,7 +1,7 @@
 cron:
 - description: update Discovery documents in googleapis/discovery-artifact-manager
   url: /cron/discoveries
-  schedule: every 30 mins synchronized
+  schedule: every 12 hours synchronized
```
vchudnov-g: Any particular reason for this new interval? Would 6hrs or 3hrs be good so that when a team publishes an API, they can see client library changes in the same workday?

saicheems: It was an arbitrary choice, just so commits aren't spammed to dartman all day. Changed to 3 hours.

PR #75, opened 2017-10-19
Copy a snapshot of toolkit

saicheems:
Preserve the implementation of Discovery samplegen.

The contents of github.com/googleapis/toolkit/ as of 2017-10-20 12:40
are captured in /toolkit/. This is a one-time sync to cut development
costs on the Toolkit repo, which will continue to evolve. We expect few
or no changes to the Discovery samplegen code, which will henceforth
only live in this repo, not in Toolkit.


saicheems:
Used your description with a small edit.

PR #76, opened 2017-11-16
Fix Go push

saicheems:
The repo now requires the flag "-o nokeycheck" to be passed to the "git
push" command.


server/tasks/google_api_go_client.py
```diff
@@ -27,42 +27,37 @@
...
+
+    Args:
+        filepath (str): the directory to work in.
+        github_account (GitHubAccount): the GitHub account to commit with.
+    """
+    env = os.environ.copy()
+    env['GOPATH'] = filepath
+    check_output(['go', 'get', '-d', '-t', 'google.golang.org/api/...'],
+                 env=env)
+    repo = _git.Repository(join(filepath, 'src/google.golang.org/api'))
```
vchudnov-g: why did this repo change from google-api-go-generator?

saicheems: That's the repo that contains `google-api-go-generator` along with all the client libraries:

https://godoc.org/google.golang.org/api

saicheems: Calling the generator updates all the APIs in that package.

PR #77, opened 2018-04-13
Add artman configs

andreamlin:
Copy over yamls for discogapic from googleapis/googleapis. Remove unused config values, like grpc and proto fields.
Calling artman on Compute from this repo successfully generates the Compute java library (assuming https://github.com/googleapis/gapic-generator/pull/1937 is patched into toolkit)


andreamlin:
PTAL

.gitignore
```diff
@@ -1,2 +1,3 @@
 *~
 *.pyc
+artman-genfiles
```
vam-google: Do you really need this? It seems like a default directory where artman will generate its stuff. Correctly used artman will not generate there (in most workflows the files will go either into specified external dir or in automatically generated temp dir).

gapic/lang/common.yaml
```diff
@@ -0,0 +1,7 @@
+common:
+  toolkit_path: ${TOOLKIT}
+  package_dependencies_yaml: ${DISCOVERY_ARTIFACT_MANAGER}/gapic/packaging/dependencies.yaml
+  package_defaults_yaml: ${DISCOVERY_ARTIFACT_MANAGER}/gapic/packaging/api_defaults.yaml
+java:
+  discogapic_language_yaml:
+    - ${DISCOVERY_ARTIFACT_MANAGER}/gapic/lang/java_discogapic.yaml
```
vam-google: Please add a new line

gapic/packaging/api_defaults.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+author: Google LLC
+email: googleapis-packages@google.com
+homepage: https://github.com/googleapis/discovery-artifact-manager
+license: Apache-2.0
+release_level:
+  python: alpha
+  java: alpha
+  ruby: alpha
+  nodejs: alpha
```
vam-google: what about Go, C# and PHP? Or, since only java is supported for now, I belive to have only java config (the less confusing an misleading config we have the better, here having configs for pyhon, ruby and node makes an impression that they are supported for discovery).

gapic/packaging/dependencies.yaml
```diff
@@ -0,0 +1,31 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Configures the version of the dependencies of various generated packages, and
+# related information.
+
+#TODO(andrealin): set up copybara from internal
```
vam-google: is it still valid todo?

gapic/lang/java_discogapic.yaml
```diff
@@ -0,0 +1,5 @@
+type: com.google.api.codegen.ConfigProto
```
vam-google: So what suffix are we using "discogapic" or "gapic"? Here it is discogapic but for compute it is "gapic". Is it intentional? Can we just use one suffix?

.gitignore
andreamlin: Ok, this does seem like tight coupling with the artman source code, which uses `artman-genfiles` for the default output directory within googleapis or wherever artman is called.

Removed.

gapic/lang/common.yaml
andreamlin: Done.

gapic/packaging/api_defaults.yaml
andreamlin: Sure; removed the other languages.

gapic/packaging/dependencies.yaml
andreamlin: Removed it. 

I'll just document in the playbook that this is the source of truth (as opposed to google3) and devs should directly edit this file.

gapic/lang/java_discogapic.yaml
andreamlin: as discussed offline, i'll leave this as is.

andreamlin:
PTAL; addressed comments.

gapic/packaging/dependencies.yaml
vam-google: Is it ok to use github as source of truth? (seems like a violation of single sourse of truth principle).

PR #81, opened 2018-05-15
Regenerate compute_gapic.yaml

andreamlin:
Regenerate the compute config using artman.

### DIFF
--- a/gapic/google/compute/v1/compute_gapic.yaml
+++ b/gapic/google/compute/v1/compute_gapic.yaml
@@ -4574,11 +4574,13 @@ interfaces:
       - parameters:
         - instance
         - requestId
+        - forceAttach
         - attachedDiskResource
     # TODO: Configure which fields are required.
     required_fields:
     - instance
     - requestId
+    - forceAttach
     - attachedDiskResource
     request_object_method: true
     resource_name_treatment: STATIC_TYPES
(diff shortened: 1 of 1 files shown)


andreamlin:
PTAL

gapic/google/compute/v1/compute_gapic.yaml
```diff
@@ -7515,6 +7517,416 @@ interfaces:
...
+  methods:
+  - name: compute.regionDisks.createSnapshot
+    # TODO: Configure which groups of fields should be flattened into method
+    # params.
+    flattening:
+      groups:
+      - parameters:
+        - disk
+        - requestId
+        - snapshotResource
```
pongad: What is this? I don't see it in the discovery doc; is it the request body?

gapic/google/compute/v1/compute_gapic.yaml
```diff
@@ -7515,6 +7517,416 @@ interfaces:
...
+    # TODO: Configure which fields are required.
+    required_fields:
+    - disk
+    - requestId
```
pongad: `requestId` is optional.

gapic/google/compute/v1/compute_gapic.yaml
```diff
@@ -7515,6 +7517,416 @@ interfaces:
...
+        - requestId
+        - sourceImage
+        - region
+        - diskResource
+    # TODO: Configure which fields are required.
+    required_fields:
+    - requestId
+    - sourceImage
+    - region
+    - diskResource
```
pongad: requestId and sourceImage are optional. This seems to be happening in other places as well.

Is "region" here the whole "projects/P/regions/R"?

gapic/google/compute/v1/compute_gapic.yaml
andreamlin: gapic-generator automatically appends `Resource` to the variable name of a request body object, in this case `snapshot`.

gapic/google/compute/v1/compute_gapic.yaml
andreamlin: Hmm this sounds like gapic-generator needs to be changed to exclude query parameters from the required_fields list config. I will make a PR for that first, which blocks this one.

gapic/google/compute/v1/compute_gapic.yaml
andreamlin: Will make query params optional in a bit.

re: "region", yes. See line 7829 for the mapping of the `region` param to its field_name_pattern `projectRegion`, which has a path template defined in lines 7532-7533.

gapic/google/compute/v1/compute_gapic.yaml
pongad: Makes sense

gapic/google/compute/v1/compute_gapic.yaml
pongad: Right makes sense too

pongad:
I'm good with this, assuming we make query params optional at some point.

PR #88, opened 2018-07-12
Removing deprecated GAPIC files

garrettjonesgoogle:


garrettjonesgoogle:
That target is still broken because of https://github.com/googleapis/artman/blob/master/artman/config/converter.py#L40 , right?

andreamlin:
Ahh yes.

andreamlin:
Ok, I used this config and added a `src_proto_paths` line and `proto_deps` line and the artman generation passed.

PR #90, opened 2018-08-09
Regenerate compute_gapic.yaml

andreamlin:
Ran `artman --config=gapic/google/compute/artman_compute.yaml --root-dir=`pwd` --local generate discogapic_config`
on latest compute.v1.json.

Successfully generated a client from this config afterwards.


andreamlin:
PTAL

PR #91, opened 2018-09-05
Regenerate compute_gapic.yaml

andreamlin:
We need to regenerate the gapic config for compute in order to refresh the library itself, because the gapic config defines new resource types and methods that were introduced in the compute Discovery doc.


andreamlin:
PTAL
The regenerated compute library passes tests https://github.com/GoogleCloudPlatform/google-cloud-java/pull/3642

PR #92, opened 2018-09-27
Bazel-based codegen #5 (discovery-artifact-manager integration)

vam-google:
This PR depends on https://github.com/googleapis/gapic-generator/pull/2326, which defines bazel rules used here.

Using bazel gapic rules to generate compute discogapic. Currently only compute client is configured.

Bazel `0.16.0+` required (with `--experimental_enable_repo_mapping` flag added to each command). 

To enable experimental feature in IntelliJ IDEA bazel plugin (and eliminate errors on sync, note this step is needed only if you try to run the commands from IntelliJ bazel plugin, and **can be skipped** if you use bazel from command line only) add the following lines to the `.bazelproject` file (generated by the plugin):

```yaml
build_flags:
  --experimental_enable_repo_mapping
```
Please see the header comment in `WORKSPACE` file (part of this PR) for justification of using experimental feature in this PR.

**Prerequisites**
1. Download bazel (the following commands assume that bazel binary is downloaded in this repository root)
```
wget -O bazel https://github.com/bazelbuild/bazel/releases/download/0.17.1/bazel-0.17.1-linux-x86_64
chmod 755 bazel
```
Just downloading the bazel binary should be enough for most environments. If that was not sufficient, please check [bazel install instructions](https://docs.bazel.build/versions/master/install-ubuntu.html).

**Usage** (each of the commands can be executed independently from each other)
1. Note the first build may take a while (because bazel will build protoc from sources), but all the subsequent build will be fast (no need to rebuild protoc).

1. To generate and build (by bazel itself, no gradle involved) the `compute/v1` client: 
```
./bazel build --experimental_enable_repo_mapping //gapic/google/compute/v1:compute_java_gapic
``` 
2. To run all generated tests for compute client (by bazel itself, no gradle involved):
```
./bazel test --experimental_enable_repo_mapping //gapic/google/compute/v1:all
``` 
3. To generate code, tests and run generated tests in all packages (currently is equal to running only compute tests, because it is the only configured client):
```
./bazel test --experimental_enable_repo_mapping //... 
```
4. To generate discogapic opensource gradle package (artman-like):
```
./bazel build --experimental_enable_repo_mapping //gapic/google/compute/v1:google-cloud-compute-v1-java
```
Check the generted `build.gradle` inside `discogapic-java-assembly/gapic-google-cloud-compute-v1-java` (in generated `.tar.gz` package). The file defines explicitly all dependencies (including transitive ones). This is a result or bazel not supporting transitive dependencies (even though gradle does). I consider this as a benefit: no hidden surprises and it is shown how many dependencies there are and what they are. This can be removed by applying whitelist in the dependencies (partially done, can be useful, if explicit definition of transitive deps becomes an issue somehow). Another option to provide a comment near each dependency if it is a transitive one or direct one (not implemented now). 

5. To build and run tests from gradle (to do it by bazel itself see the items 2 and 3):
    1. Unpack generated gradle package (see item 4):
    ```
    tar -xzvpf bazel-bin/gapic/google/compute/v1/google-cloud-compute-v1-java.tar.gz
    ```
    2. Go to unpacked `google-cloud-compute-v1-java` and run:
    ```
    ./gradlew clean test
    ```


WORKSPACE
```diff
@@ -0,0 +1,97 @@
+# IMPORTANT INFORMATION!
+#
+# This workspace configuraiton assumes `--experimental_enable_repo_mapping` command line argument
+# on every build of anything under this workspace. The repo mapping is a recently released feature
+# (bazel >= 0.16.0) and allows to solve two very nasty problems:
+#     1) Allowing two or more workspaces with dependency conflicts (same dependency name but
+#        different version) to coexist under same project (when one workspace imports another via
+#        `*_reporitory()` workspace rules). Example: `gapic-generator` and `grpc-java` have a guava
```
andreamlin: spelling: _repository

WORKSPACE
```diff
@@ -0,0 +1,97 @@
...
+#         dependency conflict (grpc-java uses latest java7 compatible version, while gapic-generator
+#         uses fresher (java8 compatible) version)
+#     2) Allowing two or more workspaces to have same dependency, but assign different names to it
+#       (i.e. at least one of the workspaces does not follow naming conventions).
+#
+# Even if the experimental feature will be eventually cancelled, there will still be a solution
+# for solving the problems desribed above, so it is relatively safe to structure workspaces and
+# packages under assumption that the dependencies naming conflicts can be solved in the "outer"
+# workspace (the one, which imports the "innter" workspace via `*_repository` rule).
```
andreamlin: spelling: inner

WORKSPACE
```diff
@@ -0,0 +1,97 @@
...
+#
+# Note, the rule arguments related to the repo mapping feature (like repo_mapping arument in
```
andreamlin: spelling: argument

WORKSPACE
```diff
@@ -0,0 +1,97 @@
...
+# `*_repository` rules) are currently highlighted as errornous by IDE plugins (is expected taking
```
andreamlin: spelling: erroneous

WORKSPACE
vam-google: It seems correct "the rule arguments", meaning the arguments of the rule, which are releated to the repo mapping feature

WORKSPACE
andreamlin: just to make sure, is it spelled "repository" instead of "reporitory"?

WORKSPACE
andreamlin: "repo_mapping arument" -> "repo_mapping ar**g**ument"

WORKSPACE
```diff
@@ -0,0 +1,97 @@
+# IMPORTANT INFORMATION!
+#
+# This workspace configuraiton assumes `--experimental_enable_repo_mapping` command line argument
```
garrettjonesgoogle: spelling (configuraiton)

WORKSPACE
```diff
@@ -0,0 +1,97 @@
...
+#       (i.e. at least one of the workspaces does not follow naming conventions).
+#
+# Even if the experimental feature will be eventually cancelled, there will still be a solution
+# for solving the problems desribed above, so it is relatively safe to structure workspaces and
+# packages under assumption that the dependencies naming conflicts can be solved in the "outer"
+# workspace (the one, which imports the "inner" workspace via `*_repository` rule).
+#
+# Note, the rule arguments related to the repo mapping feature (like repo_mapping arument in
+# `*_repository` rules) are currently highlighted as erroneous by IDE plugins (is expected taking
+# into account that the feature is new and experimental). This problem is temprorary and will go
```
garrettjonesgoogle: spelling (temprorary)

WORKSPACE
```diff
@@ -0,0 +1,97 @@
...
+# gapic-generator repository dependencies (required to compile and run gapic-generator)
+#
+load(
+    "@com_google_api_codegen//:repositories.bzl",
+    "com_google_api_codegen_repositories",
+    "com_google_api_codegen_test_repositories",
+    "com_google_api_codegen_tools_repositories",
+)
+
+#TODO:  Update all ommited dependencies in gapic_generator so they match the newer versions used by
```
garrettjonesgoogle: spelling (ommited -> omitted)

gapic/google/compute/v1/BUILD.bazel
```diff
@@ -0,0 +1,92 @@
...
+    "com.google.cloud.compute.v1.TargetHttpsProxyClientTest",
+    "com.google.cloud.compute.v1.TargetInstanceClientTest",
+    "com.google.cloud.compute.v1.TargetPoolClientTest",
+    "com.google.cloud.compute.v1.TargetSslProxyClientTest",
+    "com.google.cloud.compute.v1.TargetTcpProxyClientTest",
+    "com.google.cloud.compute.v1.TargetVpnGatewayClientTest",
+    "com.google.cloud.compute.v1.UrlMapClientTest",
+    "com.google.cloud.compute.v1.VpnTunnelClientTest",
+    "com.google.cloud.compute.v1.ZoneClientTest",
+    "com.google.cloud.compute.v1.ZoneOperationClientTest",
```
garrettjonesgoogle: We can't use some kind of glob for this?

vam-google: Unfortunately no, because glob() searches the local filesystem, but here the tests are not on filesystem, they are the output of the generator rule target. I agree that automating the tests list generation is important and must be done. I just think that at this point it is ok to keep them manually and later add the functionality, since it is not important for the `rules_gapic` core.

PR #96, opened 2018-11-02
Regen compute_gapic.yaml from gapic-generator:master

andreamlin:
Just regenerated compute_gapic.yaml from gapic-generator. Apparently resource_object_method config has become unsupported since last regen.


andreamlin:
PTAL

PR #97, opened 2018-11-06
Update common-protos dep to match gapic-generator

andreamlin:
gapic-generator just started using v1.13.0-pre2 in https://github.com/googleapis/gapic-generator/pull/2394


andreamlin:
PTAL

PR #100, opened 2018-12-10
Regenerate compute_gapic.yaml

andreamlin:
Regenerated with toolkit/master.

See https://github.com/googleapis/google-cloud-java/pull/4206 for the resulting generated client.


andreamlin:
PTAL

PR #101, opened 2018-12-10
Sample regen of compute, no manual edits (DO NOT MERGE)

andreamlin:
Example of the result when `artman --config=gapic/google/compute/artman_compute.yaml  generate discogapic_config` is used to generated the `compute_gapic.yaml` file.

This contains no hand edits. Therefore, the diff in this PR are things that I have to manually revert back to the right values, every time I regenerate this file..

### DIFF
--- a/gapic/google/compute/v1/compute_gapic.yaml
+++ b/gapic/google/compute/v1/compute_gapic.yaml
@@ -8,7 +8,7 @@ config_schema_version: 1.0.0
 # The settings of generated code in a specific language.
 language_settings:
   java:
-    package_name: com.google.cloud.compute.v1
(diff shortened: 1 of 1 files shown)


gapic/google/compute/v1/compute_gapic.yaml
```diff
@@ -8,7 +8,7 @@ config_schema_version: 1.0.0
 # The settings of generated code in a specific language.
 language_settings:
   java:
-    package_name: com.google.cloud.compute.v1
+    package_name: com.google.compute.v1
```
andreamlin: This is a regression.

PR #104, opened 2018-12-17
Make toolkit compile

hzyi-google:
Commit "Copy a snapshot of toolkit" 958f34ec9bc4ac1aa883e35e0137382e784c1319 does not seem to be an exact copy of toolkit as of 2017-10-20 12:40. The toolkit in this repo does not compile.

This commit is created by running `git reset --hard e4541cdfd24cbe3d6e0bc763aa7d7d3c32f010d1` and `git clean -d -f`. The commit hash points to the commit "Ruby: Disambiguate mock credentials classes in tests. (#1644)" in toolkit, which is right before the commit where discovery samplegen is deleted.

This commit will revert "change nodejs import to const {google}", which will be fixed in a coming PR.


hzyi-google:
@vchudnov-g Can you double check and make sure:
1. the toolkit in master branch does not compile (you may need to upgrade the version of `protobuf-gradle-plugin` to above 0.8.0 beforehand. 
2. the toolkit in this branch compiles and all tests pass after running `gradle clean install` and `gradle test`?

PR #105, opened 2018-12-19
get back the commit change nodejs import to const {google}

hzyi-google:
The commit was reverted in #104. This pull request gets it back.


hzyi-google:
PTAL

PR #106, opened 2019-01-10
Update the nodejs sample templates

JustinBeckwith:


JustinBeckwith:
👋 @hzyi-google - I made this change just to get the ball rolling, but I have *no* idea how to test it.  Can you help me figure out how to run this thing and check the output? Or are there tests that should be running in Travis?

hzyi-google:
Hi @JustinBeckwith , you can run `gradle test` from toolkit/ to test the templates against baseline. It will compare the samples generated from the templates with baselines [here](https://github.com/googleapis/discovery-artifact-manager/tree/master/toolkit/src/test/java/com/google/api/codegen/testdata/discoveries/nodejs).
That said, you also need to update those baseline files.

hzyi-google:
It seems `compute` samples are treated as `js` instead of `nodejs`, so [this](https://github.com/googleapis/discovery-artifact-manager/blob/master/toolkit/src/main/resources/com/google/api/codegen/js/sample.snip) probably needs to be updated too.

PR #112, opened 2019-03-26
Increase memory allocation

dwsupplee:
Go apiary builds are continuing to be fail due to running out of memory. Nudging this value up a bit to hopefully resolve this in the short term, but we may want to look into a more permanent solution.

### DIFF
--- a/server/app.yaml
+++ b/server/app.yaml
@@ -3,4 +3,4 @@ env: flex
 automatic_scaling:
   max_num_instances: 2
 resources:
-  memory_gb: 3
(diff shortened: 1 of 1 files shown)


saicheems:
Drive by comment, but maybe [limiting the set of packages under test](https://github.com/googleapis/discovery-artifact-manager/blob/master/server/tasks/google_api_go_client.py#L59), or even skipping the testing step may be the right answer at this point. It looks like the google-api-go-client package got too big to compile in its entirety?

PR #115, opened 2019-04-16
Re-enable cloud health care generation for PHP/ruby

dwsupplee:
We had previously halted generation of the Cloud HealthCare clients due to issues with `$`s in method names. A fix is being pushed to the service which will resolve this.

Please hold off on merging until we get the ok the service is updated in production.


dwsupplee:
The service has been updated, this should be ready.

PR #116, opened 2019-05-01
Healthcare: Allow method names to be in Dash-case

hzyi-google:
Healthcare discovery doc has three method names that need to conform to external conventions and cannot be in camel case(e.g., `Observation-lastn`). This PR catches this behavior and converts them to camelCase before sending them to samplegen.


hzyi-google:
PTAL

toolkit/src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -190,6 +212,30 @@ private static boolean isCamel(String identifier, CheckCase check) {
     return true;
   }
 
+  public static boolean isHyphen(String identifier, CheckCase check) {
```
dwsupplee: There are a few mentions of dash-case in this addition, WDYT About calling this and the other related `hyphen` methods `isDash`, etc.?

Also, after a quick glance it looks like most of the public methods in this utility class have docblocks - would it make sense to add one here too?

toolkit/src/main/java/com/google/api/codegen/util/Name.java
```diff
@@ -190,6 +212,30 @@ private static boolean isCamel(String identifier, CheckCase check) {
...
+    }
+    String[] subPieces = identifier.split("-");
+    if (subPieces.length <= 1) {
+      return false;
+    }
+    for (String subPiece : subPieces) {
+      if (!check.valid(identifier.charAt(0))) {
+        return false;
+      }
+      if (!subPiece.chars().skip(1).allMatch(Character::isLowerCase)) {
```
dwsupplee: If possible, it would be great to see some test cases [here](https://github.com/googleapis/discovery-artifact-manager/blob/7df9e97538f976c2483fc7a04c6b8f3795419d6c/toolkit/src/test/java/com/google/api/codegen/util/NameTest.java) exercising this logic.

toolkit/src/main/java/com/google/api/codegen/util/Name.java
hzyi-google: I chose `hyphen` to be consistent with guava ([com.google.common.base.CaseFormat](https://google.github.io/guava/releases/20.0/api/docs/com/google/common/base/CaseFormat.html)) - but I think being consistent within this repo is important as well. How about changing other mentioning of `dash-case` to `hyphen-case`?

dwsupplee: Sounds good!

hzyi-google:
PTAL @dwsupplee

PR #117, opened 2019-06-14
fix: example of fetching auth client does not match API

bcoe:
Our authentication API is now promise based, and has in general a slightly different API signature than documented in these historic samples.

### DIFF
--- a/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -165,16 +165,12 @@
 
 @private authFunc_ADC(class)
   function {@class.auth.authFuncName}(callback) {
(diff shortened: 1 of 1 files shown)


vchudnov-g:
Did you run the tests? I believe for those to pass you need to make the corresponding change to the baseline file.
(we haven't disabled the tests, have we @hzyi-google ?)

JustinBeckwith:
Apologies - saw passing CI and kind of assumed that would be running the tests - is that not the case?

hzyi-google:
I don't think this repo runs tests on PRs... And yes baselines need to be updated as well. I can submit a follow-up PR.

vchudnov-g:
I don't see CI configured in CircleCI. It's quite possible we never got this repo set up with them yet (maintenance mode and all that). @hzyi-google would have the current information on how to run the tests manually. (It's not obvious because this repo includes a frozen snapshot of gapic-generator)....

PR #119, opened 2019-06-17
[fix] correctly merge snippet files

hzyi-google:
- change temporary file permissions to allowing writing to those files
- sheet and storage have new languages: "web". We don't need to generate samples for them, but we should be able to merge them.


hzyi-google:
PTAL

hzyi-google:
@vchudnov-g friendly ping

hzyi-google:
The permissions cause problems when merging in manual samples.

PR #137, opened 2020-02-10
refactor(sample): switch to async/await which is closer to other samples

bcoe:
switches to `async`/`await` from promise chain, which is closer to other code samples, and looks less confusing than promises being used in association with callbacks.

### DIFF
--- a/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -164,14 +164,16 @@
 @end
 
 @private authFunc_ADC(class)
-  function {@class.auth.authFuncName}(callback) {
-    google.auth.getClient({
+  async function {@class.auth.authFuncName}(callback) {
+    const auth = new google.auth.GoogleAuth({
       scopes: ['https://www.googleapis.com/auth/cloud-platform']
-    }).then(client => {
-      callback(client);
-    }).catch(err => {
-      console.error('authentication failed: ', err);
     });
+    try {
+      const authClient = await auth.getClient();
+      callback(authClient);
+    } catch (err) {
+      console.error('authentication failed: ', err);
+    }
   }
 @end
 
--- a/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -20,7 +20,7 @@
   @end
 
   const {{@class.googleImportVarName}} = require('googleapis');
-  var {@class.serviceVarName} = {@class.googleImportVarName}.{@class.serviceTypeName}('{@class.apiVersion}');
+  const {@class.serviceVarName} = {@class.googleImportVarName}.{@class.serviceTypeName}('{@class.apiVersion}');
 
   @switch class.auth.type
   @case "NONE"
--- a/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -191,9 +191,9 @@
 
     if (authClient == null) {
       throw Error('authentication failed');
-    } else {
-      return authClient;
     }
+
+    return authClient;
   }
 @end
 
--- a/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
+++ b/toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
@@ -41,9 +41,11 @@
 @end
 
 @private mainFunc(class)
-  {@class.auth.authFuncName}(function({@class.auth.authVarName}) {
+  async function main () {
+    const {@class.auth.authVarName} = await {@class.auth.authFuncName}();
     {@mainFuncBody(class)}
-  });
+  }
+  main();
 @end
 
 @private mainFuncBody(class)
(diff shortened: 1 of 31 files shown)


toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
```diff
@@ -164,14 +164,16 @@
 @end
 
 @private authFunc_ADC(class)
-  function {@class.auth.authFuncName}(callback) {
-    google.auth.getClient({
+  async function {@class.auth.authFuncName}(callback) {
```
JustinBeckwith: Thinking that the async function probably doesn't take a callback, right?

toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
```diff
@@ -164,14 +164,16 @@
...
+    const auth = new google.auth.GoogleAuth({
       scopes: ['https://www.googleapis.com/auth/cloud-platform']
-    }).then(client => {
-      callback(client);
-    }).catch(err => {
-      console.error('authentication failed: ', err);
     });
+    try {
```
JustinBeckwith: Thinking we shouldn't wrap with the try/catch here, and just let it throw.

toolkit/src/main/resources/com/google/api/codegen/nodejs/sample.snip
```diff
@@ -191,10 +190,10 @@
     var authClient = null;
 
     if (authClient == null) {
-      console.log('authentication failed');
-      return;
+      throw Error('authentication failed');
+    } else {
```
JustinBeckwith: nit: we don't need the else because the if throws

toolkit/src/test/java/com/google/api/codegen/testdata/discoveries/nodejs/nodejs_adexchangebuyer.v1.4.json.baseline
```diff
@@ -8,9 +8,10 @@
 //    `npm install googleapis --save`
 
 const {google} = require('googleapis');
-var adexchangebuyer = google.adexchangebuyer('v1.4');
+const adexchangebuyer = google.adexchangebuyer('v1.4');
 
-authorize(function(authClient) {
+async function main () {
+  const authClient = await authorize();
```
JustinBeckwith: I hate to do this to you - but now this full snippet doesn't make sense.  Keep looking below here - we are now calling `adexchangebuyer.accounts.get`, and passing a callback, but inside of an async function.  Further blow, there is an `authorize` function that does not make any sense that I can tell.  

I know this isn't your mess, but it would probably be a good idea to pop open the whole baseline, and just make sure it makes sense.

bcoe: grumble grumble, you can have callbacks and functions, grumble grumble.

Looking, I agree that `async`/`await` would be more idiomatic.

bcoe: I undertook a larger refactor, I think finding a bug in the process (which is that we actually want to access `data` on the response).

Your example:

```js
const {google} = require('googleapis');
const adexchangebuyer = google.adexchangebuyer('v1.4');

async function main () {
  const authClient = await authorize();
  const request = {
    // The accountId of the publisher to get profiles for.
    accountId: 0,  // TODO: Update placeholder value.

    auth: authClient,
  };

  try {
    const response = (await adexchangebuyer.pubprofiles.list(request)).data;
    // TODO: Change code below to process the `response` object:
    console.log(JSON.stringify(response, null, 2));
  } catch (err) {
    console.error(err);
  }
}
main();

async function authorize() {
  // TODO: Change placeholder below to generate authentication credentials. See
  // https://foo.com/bar
  //
  // Authorize using the following scope:
  //   'https://www.googleapis.com/auth/adexchange.buyer'
  let authClient = null;

  if (authClient == null) {
    throw Error('authentication failed');
  }

  return authClient;
}
```

A more complex paginated example:

```js
const {google} = require('googleapis');
const compute = google.compute('v1');

async function main () {
  const authClient = await authorize();
  const request = {
    // Project ID for this request.
    project: 'my-project',  // TODO: Update placeholder value.

    auth: authClient,
  };

  let response;
  do {
    if (response && response.nextPageToken) {
      request.pageToken = response.nextPageToken;
    }
    response = (await compute.zones.list(request)).data;
    const itemsPage = response.items;
    if (itemsPage) {
      for (let i = 0; i < itemsPage.length; i++) {
        // TODO: Change code below to process each resource in `itemsPage`:
        console.log(JSON.stringify(itemsPage[i], null, 2));
      }
    }
  } while (response.nextPageToken);
}
main();

async function authorize() {
  const auth = new google.auth.GoogleAuth({
    scopes: ['https://www.googleapis.com/auth/cloud-platform']
  });
  return await auth.getClient();
}
```

an example using an apiKey.

```js
const {google} = require('googleapis');
const translate = google.translate('v2');

const apiKey = '';  // TODO: Update placeholder with desired API key.

async function main () {
  const request = {
    // The text to translate
    q: '',  // TODO: Update placeholder value.

    // The target language into which the text should be translated
    target: '',  // TODO: Update placeholder value.

    auth: apiKey,
  };

  try {
    const response = (await translate.translations.list(request)).data;
    // TODO: Change code below to process the `response` object:
    console.log(JSON.stringify(response, null, 2));
  } catch (err) {
    console.error(err);
  }
}
main();
```

PR #141, opened 2020-03-28
fix: Fix bazel generation for discogapic clients.

vam-google:
Also update `compute_gapic.yaml`.

Note, the checked in `cumpute_gapic.yaml` is not used for bazel generation and is kept here only for backward compatibility with artman generation. Bazel generates the gapic file on the fly.


hzyi-google:
LGTM

PR #143, opened 2020-04-14
fix Dockerfile so it can build

### LINKED ISSUE #142: Dockerfile no longer builds
#### Environment details

  - OS: macos

#### Steps to reproduce

1. Clone the repo
2. `cd server`
3. `docker build .`

Will fail on the follow step: `RUN virtualenv /env -p python3`

codyoss:
Fixes #142


codyoss:
@dazuma Thanks, can you merge too. I don't have write access.

PR #145, opened 2020-04-21
refactor Go generation

codyoss:
There seems to be something strange where GO111MODULE=on is not
being honored by the `make all` command. I believe this is because
the code was in GOPATH.

Removed the need for GOPATH by making use of git to download the
src code.


codyoss:
@dazuma Could you review another in this repo for me 😸

PR #147, opened 2020-05-20
chore: remove go updater

codyoss:
With the recent change of having google-api-go-client's source of
truth be GitHub we decided to re-work the automation. The nightly
build of the discovery docs will now happen via autosynth.


codyoss:
Closing this PR and making a new one that fully cleans things up.

PR #151, opened 2020-06-05
Discovery samples: Ensure the last element of the Java import path is capitalized

vchudnov-g:
This was causing problems for IAM, since the Discovery file said `"canonicalName" : "iam"` and thus the import in the sample read `import ....v1.iam;` instead of `import ....v1.Iam;`

### DIFF
--- a/toolkit/src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
+++ b/toolkit/src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
@@ -73,7 +73,15 @@ class JavaSampleTypeNameConverter implements SampleTypeNameConverter {
 
   @Override
   public TypeName getServiceTypeName(String apiTypeName) {
-    return typeNameConverter.getTypeName(Joiner.on('.').join(packagePrefix, apiTypeName));
+    return typeNameConverter.getTypeName(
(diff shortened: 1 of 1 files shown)


toolkit/src/main/java/com/google/api/codegen/discovery/transformer/java/JavaSampleTypeNameConverter.java
```diff
@@ -73,7 +73,15 @@ public JavaSampleTypeNameConverter(String apiName, String apiVersion, boolean ve
 
   @Override
   public TypeName getServiceTypeName(String apiTypeName) {
-    return typeNameConverter.getTypeName(Joiner.on('.').join(packagePrefix, apiTypeName));
+    return typeNameConverter.getTypeName(
+        Joiner.on('.').join(packagePrefix, capitalize(apiTypeName)));
+  }
+
+  public static String capitalize(String str) {
```
vam-google, line 80: Should this be public (maybe private is enough, I see it being used from getSErviceTypeName()?

vchudnov-g: Done.

vchudnov-g:
Argh. Not fully fixed, it turns out. The references to the imported class are still lower-cased. I'll update this PR with a fix.

vchudnov-g:
Ok, that should be fixed now. I'll figure out how to add a test before I submit this.

PR #153, opened 2020-06-16
Replacing deprecated oauth2client with google-auth

### LINKED ISSUE #152: Python: oauth2client is deprecated and should be replaced with google-auth
**Is your feature request related to a problem? Please describe.**
https://github.com/googleapis/oauth2client is deprecated. Consequently, automatically generated code samples shouldn't continue to produce code that uses oauth2client for default credentials.

 **Describe the solution you'd like**
A simplest way would be replace fragments that use 
```python
from oauth2client.client import GoogleCredentials

credentials = GoogleCredentials.get_application_default()
```

with 
```python
from google import auth

credentials, _ = auth.default()
```
and update instructions to install *both* google-api-python-client and google-auth:
`pip install --upgrade google-api-python-client google-auth`

**Describe alternatives you've considered**
This is probably the best and the simplest solution to the oauth2client deprecation.

DFrenkel:
Fixes #152

### Notes
* discoGen jar is generated via `cd toolkit && ./gradlew discoJar`

### Testing
* `cd toolkit && ./gradlew test` passes
* I also modified run_test.py to produce override with `'authType': 'APPLICATION_DEFAULT_CREDENTIALS'` and ran `./run_tests.py -l python discoveries/cloudbilling.v1.json`. All tests also passed as soon as I generated default credentials via `gcloud auth application-default login` and all generated code samples looked sane.

### DIFF
--- a/toolkit/src/main/resources/com/google/api/codegen/py/discovery_fragment.snip
+++ b/toolkit/src/main/resources/com/google/api/codegen/py/discovery_fragment.snip
@@ -19,9 +19,9 @@
             2. This sample uses Application Default Credentials for authentication.
                If not already done, install the gcloud CLI from
                https://cloud.google.com/sdk/ and run
-               `gcloud beta auth application-default login`
-            3. Install the Python client library for Google APIs by running
-               `pip install --upgrade google-api-python-client`
+               `gcloud auth application-default login`
+            3. Install the Python client library for Google APIs and Google Auth Python Library by running
+               `pip install --upgrade google-api-python-client google-auth`
         @else
             2. Install the Python client library for Google APIs by running
                `pip install --upgrade google-api-python-client`
@@ -47,7 +47,7 @@
             @# Use Application Default Credentials for authentication when running locally.
             @# For more information, see:
             @# https://developers.google.com/identity/protocols/application-default-credentials
-            {@importGoogleCredential()}credentials = GoogleCredentials.get_application_default()
+            @# Google API Client will automatically use them as long as googl-auth is installed.
             {@BREAK}
         @case "OAUTH_3L"
             @# {@TODO()} Change placeholder below to get authentication credentials.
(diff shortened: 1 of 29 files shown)


toolkit/src/main/resources/com/google/api/codegen/py/discovery_fragment.snip
```diff
@@ -47,7 +47,7 @@
             @# Use Application Default Credentials for authentication when running locally.
             @# For more information, see:
             @# https://developers.google.com/identity/protocols/application-default-credentials
-            {@importGoogleCredential()}credentials = GoogleCredentials.get_application_default()
+            {@importGoogleAuth()}credentials, _ = auth.default()
```
busunkim96, line 50: I'd prefer we remove the call to default - the library will make this call under the hood as long as google-auth is installed.

```py
from googleapiclient import discovery

service = discovery.build("appengine", "v1beta5")
```

https://github.com/googleapis/google-api-python-client/blob/21af37b11ea2d6a89b3df484e1b2fa1d12849510/googleapiclient/_auth.py#L41-L52

This also better matches what we do in [handwritten samples](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/storage/cloud-client/storage_download_file.py#L29).

DFrenkel: Good idea, thanks - will remove explicit call to default in the next iteration of the pull request. 

discoGen jar is being rebuilt and included into the pull request intentionally - some of the code generation logic lives inside the jar. Anything not related to discovery snippet, it seems, is being generated by that Jar, and in fact discovery snippet changes above don't affect  the way most samples are generated.

DFrenkel: This suggestion is now implemented: 
some extra complexity in templating arose because now `APPLICATION_DEFAULT_CREDENTIALS` case is different from `OAUTH_3L` case as we no longer provide credentials to discovery build in the former case and need to continue doing so in the latter case. This is why there is not an update to  `toolkit/src/main/java/com/google/api/codegen/discovery/transformer/py/PythonSampleMethodToViewTransformer.java`

busunkim96:
@JustinBeckwith Could you merge this? (It looks like I don't have write access)

JustinBeckwith:
@vchudnov-g are there any manual steps required for running this, and updating the snippets on GCS?

PR #158, opened 2020-09-25
fix: first step to stop pushing commits to master

SurferJeffAtGoogle:
1. (this PR) create a /cron/discoveries-pr handler that creates a pull request instead of pushing to master.
2. Create a github action that automatically approves generated PRs.
3. Set up merge on green to merge generated PRs.


server/cron.yaml
```diff
@@ -48,3 +48,7 @@ cron:
 - description: release (force) google/google-api-ruby-client
   url: /cron/clients/ruby/release?force=true
   schedule: every 99999 hours  # 11 years.
+
+- description: create pull request to update Discovery documents in googleapis/discovery-artifact-manager
+  url: /cron/discoveries-pr
+  schedule: every 99999 hours  # 11 years.
```
chingor13, line 54: I assume after this is deployed and we test it with a manual trigger, we will switch this one to every 3 hours and turn off the other's cron (or remove).

server/tasks/discovery_artifact_manager.py
```diff
@@ -96,8 +114,33 @@ def update(filepath, github_account):
...
+    """Creates a pull request on the discovery-artifact-manager repository.
+
+    Args:
+        filepath (str): the directory to work in.
+        github_account (GitHubAccount): the GitHub account to commit with.
+    """
+    repo = _git.clone_from_github(
+        _repo_path(), join(filepath, _repo_name()), github_account=github_account)
+    branch = ('update-discovery-artifacts-' +
+        datetime.datetime.now().strftime('%Y%m%d-%H%M%S'))
```
chingor13, line 134: Do we possibly see the resulting PR not merging and needing a new branch and new PR (versus the autosynth strategy of re-using the same branch name)?

server/tasks/discovery_artifact_manager.py
```diff
@@ -96,8 +114,33 @@ def update(filepath, github_account):
...
+    repo.checkout_new(branch)
+    if _update_disco(repo, github_account):
```
chingor13, line 136: nit: I'd prefer comparing the number of commits rather than assuming 0 is falsy (or make _update_disco return a boolean), but maybe this is fine idiomatic python.

chingor13 (approved):
Sorry I missed this PR.

LGTM with a few questions

server/cron.yaml
SurferJeffAtGoogle: Yep.

server/tasks/discovery_artifact_manager.py
SurferJeffAtGoogle: I think it will be easier to detect and debug the failure case with a timestamp.

server/tasks/discovery_artifact_manager.py
SurferJeffAtGoogle: Section 2.14 of Google's Python Style guide says to "Use the "implicit" false if at all possible."  But, actually, I was using implicit true.  So, I updated the code to compare to zero.

PR #159, opened 2020-09-30
build: add GitHub Actions workflow for discovery sync

JustinBeckwith:
This is in pursuit of enabling default branch protection.  This PR does a few things:
- It introduces a `npm run update-discovery` command, which updates all discovery files locally 
- ~~It runs said script, which causes a lot of changes (which I want feedback on @vchudnov-g).  Some of the changes are sorting keys to reduce diffs in future commits.~~
- It adds an automated workflow for running this process nightly
- It adds an automated workflow for auto-approving and merging the PR

After something like this lands and is verified, the bot that auto-pushes to master would no longer be needed.  We could also enable default branch protection rules and codeowners.

### DIFF
--- a/.github/workflows/update-discovery.yml
+++ b/.github/workflows/update-discovery.yml
@@ -1,0 +1,40 @@
+name: Update Discovery
+
+on:
+  schedule:
+    - cron: "0 */3 * * *"
+  workflow_dispatch:
+
+jobs:
+  update-discovery:
+    if: ${{ github.repository_owner == 'googleapis' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install
+      - run: npm run update-discovery
+      - name: Push changes to branch
+        run: |
+          git config --global user.email "yoshi-automation@google.com"
+          git config --global user.name "Yoshi Automation Bot"
+          git checkout -b action/auto-update-discovery
+          git commit -a -m "chore: automatic update of discovery json"
+          git push -f origin action/auto-update-discovery
+      - name: Open PR
+        uses: actions/github-script@v2
+        with:
+          github-token: ${{ secrets.YOSHI_CODE_BOT_TOKEN }}
+          script: |
+            const result = await github.pulls.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              title: "chore: automatic update of discovery json",
+              head: context.repo.owner + ":action/auto-update-discovery",
+              base: "master",
+              maintainer_can_modify: true,
+              body: "This was created by a GitHub Action 🤖"
+            });
+            core.info(`Created PR ${result.data.number}`);
--- a/package.json
+++ b/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "discovery-artifact-manager",
+  "private": "true",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -0,0 +1,25 @@
...
+  "private": "true",
+  "version": "1.0.0",
+  "description": "The Discovery Artifact Manager is intended to facilitate testing, publishing, and synchronization of generators and artifacts for client libraries and generated code samples of Google APIs defined by the API Discovery Service.",
+  "main": "update-discovery.js",
+  "scripts": {
+    "update-discovery": "node update-discovery.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/JustinBeckwith/discovery-artifact-manager.git"
```
chingor13, line 12: googleapis/discovery-artifact-manager?

.github/workflows/update-discovery.yml
```diff
@@ -0,0 +1,40 @@
+name: Update Discovery
+
+on:
+  schedule:
+    - cron: "0 1 * * *"
```
chingor13, line 5: I think the current executor happens every 3 hours. A few of our autosynth jobs run around 1 or 2 AM PST as well.

.github/workflows/update-discovery.yml
```diff
@@ -0,0 +1,40 @@
...
+        run: |
+          git config --global user.email "yoshi-automation@google.com"
+          git config --global user.name "Yoshi Automation Bot"
+          git checkout -b action/auto-update-discovery
+          git commit -a -m "chore: automatic update of discovery json"
+          git push -f origin action/auto-update-discovery
+      - name: Open PR
+        uses: actions/github-script@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
chingor13, line 29: The default github-actions cannot sign the CLA.

```suggestion
          github-token: ${{ secrets.YOSHI_CODE_BOT_TOKEN }}
```

chingor13 (changes requested):
Maybe let the first opened PR make the changes?

chingor13:
Can we compile and run the existing go code that does the fetching and (refactor to node|implement sorting on the go version) in a subsequent PR?

JustinBeckwith:
@chingor13 I think I got everything here

chingor13:
[App Engine Task](https://github.com/googleapis/discovery-artifact-manager/blob/master/server/tasks/discovery_artifact_manager.py#L87-L96) which shells out to go.

[Go update binary](https://github.com/googleapis/discovery-artifact-manager/blob/master/src/main/updatedisco/main.go) 

[Go business logic](https://github.com/googleapis/discovery-artifact-manager/blob/master/src/main/common/updatedisco.go)

update-discovery.js
```diff
@@ -0,0 +1,117 @@
+// Copyright 2020 Google LLC
```
SurferJeffAtGoogle, line 1: This 100 lines of new javascript is replacing about 100 lines of python that has been running for years.  

Please add tests that confirm this new code generates the same output for the same input as the old code.  Fuzzed tests would be ideal, but I won't insist.

SurferJeffAtGoogle (changes requested):
If we merge this PR, the bot will still be needed because there are other endpoints on the bot triggered by other cron jobs.

What should we do with https://github.com/googleapis/discovery-artifact-manager/pull/158?

JustinBeckwith:
So I've come to the conclusion that this was a bad idea 😆  I was messing around with the final list of ~5 non-branch protected repos, and should have looked closer to see that @SurferJeffAtGoogle already did this in a much, much better way in #158.  Apologies for the confusion folks!

PR #173, opened 2020-10-02
chore: github action auto-approves PRs from googleapis-publisher

SurferJeffAtGoogle:


.github/workflows/autoapprove.yml
```diff
@@ -0,0 +1,41 @@
...
+      - opened
+
+jobs:
+  auto-approve:
+    runs-on: ubuntu-latest
+    steps:
+      - name: AutoApprove
+        uses: actions/github-script@v2
+        with:
+          github-token: ${{secrets.YOSHI_APPROVER_TOKEN}}
```
chingor13, line 17: secrets aren't available to PRs from forks. you may want to add an `if` condition to the job so it will skip if it's coming from a forked repo.

SurferJeffAtGoogle:
Thanks Jeff!  I will close this PR and create a new one from the same repo.

PR #176, opened 2020-10-02
chore: github action auto-approves PRs from googleapis-publisher

SurferJeffAtGoogle:


chingor13 (approved):
You may still want the [jobs.if](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif) so that if someone does send a PR from a fork, it'll show as a skip rather than an error.

PR #371, opened 2020-10-16
chore: drop old cron job that pushed to master branch

SurferJeffAtGoogle:
because the new cron job discoveries-pr creates a pull request that's automatically merged.


chingor13 (approved):

chingor13:
Note: I don't think this repo as auto deployment so the cron.yml will need to be pushed.

PR #381, opened 2020-10-19
feat: remove go updater

codyoss:
It looks like PR #147 never made it in which caused some issues for the Go client libraries when this project was last re-deployed.

Code was pushed directly to our apiary mirror which broke our automation that syncs github to googlesource. I will fix that issue once this code is merged and deployed.


codyoss:
New version of #147.

codyoss:
@SurferJeffAtGoogle Can I get a review on this. I would like to release as soon as it gets in as well. It looks like the above mentioned PR never made it in which caused some issues for the Go client libraries when this project was last re-deployed.

Code was pushed directly to our apiary mirror which broke our automation that syncs github to googlesource. I will fix that issue once this code is merged and deployed.

SurferJeffAtGoogle (approved):

PR #2074, opened 2021-06-17
Add GitHub Action to lint Python code

cclauss:
Python code is written in Python 2 which died more than 500 days ago.
https://github.com/cclauss/discovery-artifact-manager/actions


JustinBeckwith:
Thanks for this PR!  Right now this codebase is very much in maintenance mode, and we're not looking to add significant infrastructure like this adds.

Issue #6142, opened 2022-12-06
Question: .github/workflows/list-services.yml prints "index" among others. Is this correct?


suztomo:
While troubleshooting https://github.com/googleapis/google-api-java-client-services/issues/13409, I found that the job uses this repository's .github/workflows/list-services.yml for its inputs to list "services". One of them is "index", which fails always.

This repository's .github/workflows/list-services.yml prints "index" among other services. Is this "index" correct value as a service?

bshaffer:
We are getting a failing job for the service "index" because the version does not exist

https://github.com/googleapis/google-api-php-client-services/actions/runs/3753967400/jobs/6377765948

I think this is an error

PR #6315, opened 2022-12-22
fix: remove "index" from services

bshaffer:
fixes https://github.com/googleapis/discovery-artifact-manager/issues/6142

Index is not a service, so it should be excluded from the list of services.


chingor13 (approved):

bshaffer:
[it worked](https://github.com/googleapis/google-api-php-client-services/actions/runs/3761402639)!

PR #10784, opened 2024-07-01
build: add workflow to normalize discovery docs

chingor13:
A manually triggered one-off action/script to create a PR like: https://github.com/googleapis/discovery-artifact-manager/pull/10699

Future discovery docs are already being normalized in the update_disco.py script, but it won't be updated in the directory until there is a new revision with substantial changes. This will normalize all the existing docs in place (sorted keys, consistent indentation) so future diffs are readable/reviewable.

### DIFF
--- a/.github/workflows/normalize-disco.yml
+++ b/.github/workflows/normalize-disco.yml
@@ -1,0 +1,30 @@
+name: Normalize discovery documents
+on:
+  workflow_dispatch:
+
+jobs:
+  normalize-discoveries:
+    if: ${{ github.repository == 'googleapis/discovery-artifact-manager' }}
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/normalize-disco.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      - name: Install Python 3.10
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+      - name: execute
+        run: |
+          python scripts/normalize_discovery.py
```
codyoss, line 20: Looks like the python file is empty

chingor13: oops, didn't save the buffer (copied from internal)

codyoss (approved):

PR #10806, opened 2024-08-02
chore: remove google-api-client-generator

chingor13:
The per-language generators have been in per-language repositories for years. This version has not been touched in 7+ years.


razvanphp:
Why is this not merged? It's very confusing indeed and does not encurage community contributions. Thank you!

